PR #224, opened 2021-06-15
Sort parameter values and variables in ascending order 

RosaRomeroGomez:
**What**
- Sorted parameter values in ascending order
- Sorted variables panel in ascending order
- I also cleaned up some unused code in the `Simulation` view that has been carried over from copying and pasting.
- BONUS: I also added the counter values for the variables panel. 

**How**
- Used the `orderBy` lodash function.
- For the variables counters, I mostly emulated how we do it for the parameters counter.

**Screenshot**
image

**Notes**
In the future, we might want to add support for the user to select the order criteria (descending, ascending, etc.)

**Testing**
- Select a model.
- Click on `Open Simulation View`
- Parameters and variables should be ordered in ascending order.

### DIFF
--- a/client/src/views/Simulation/Simulation.vue
+++ b/client/src/views/Simulation/Simulation.vue
@@ -90,7 +90,6 @@
 
   import VariablesPanel from './components/VariablesPanel.vue';
 
-
   /**
   Temporary hack for workshop
   **/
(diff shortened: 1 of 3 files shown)


YohannParis (approved):
Thank you!

PR #225, opened 2021-06-16
Run button model simulation

### LINKED ISSUE #216: Create Run component for Model Simulation
- Call the Donu API to run a model based on current parameters, and display the results.

Screen Shot 2021-05-27 at 17 12 01

YohannParis:
**What**
- fix #216
- Create a Run button component to handle simulation run.
- Also separated the model panel off the simulation view on its own component.

**How**
- Added parameters to Donu _Services_, _Store_, and _Types_ to handle the Run configurations `{ end; start; stop }`.
- Move the call to request a run out of the variables panel into the simulation view, and renamed it `fetchResults` to match the parameters call.
- Added a `fetchVariables` to display the Variables before a run.
- Created a new Run button component to allow change of run configuration.
- Renamed the parameters container to a panel to match the other panels.
- Fixed counters information on all panels.
- Added missing types around.
- Created a setting dropdown button for the model panel component.

**Screenshot**

test

**Enhancement**
The simulation control bar is too high and needs to be cleaned up to be responsive and _good_ looking.


client/src/views/Simulation/Simulation.vue
```diff
@@ -1,74 +1,53 @@
...
+        </button>
+        <button class="btn btn-primary m-1">
+          <font-awesome-icon :icon="['fas', 'project-diagram' ]" />
+          <span>Provenance Graph </span>
+        </button>
+        <button class="btn btn-primary m-1" @click="fetchParameters">
+          <span>Reset</span>
+        </button>
+        <button class="btn btn-primary m-1" @click="incrParametersMaxCount">
+          <span>Save</span>
```
RosaRomeroGomez, line 19: These 2 buttons have to do with saving the experiment and resetting the experiments to the default configuration. I think we should move them closer to the `Run` button and maybe we can replace them with just icons to save some space? Since they are not as important we can justify that decision.

client/src/views/Simulation/components/ModelPanel.vue
```diff
@@ -0,0 +1,122 @@
...
+      <counters slot="left" :title="modelName" :data="countersData" />
+      <aside slot="right">
+        <div class="dropdown">
+          <button
+            class="btn btn-primary dropdown-toggle"
+            type="button"
+            @click="settingsOpen = !settingsOpen">
+            Settings
+          </button>
+          <div class="dropdown-menu dropdown-menu-right" :class="{ 'show': settingsOpen }">
```
RosaRomeroGomez, line 13: Maybe we want to make these settings panel closer to how we do it in the `Model` view:
image

client/src/views/Simulation/components/ModelPanel.vue
```diff
@@ -0,0 +1,122 @@
+<template>
+  <section class="simulation-model-panel">
+    <settings-bar>
+      <counters slot="left" :title="modelName" :data="countersData" />
+      <aside slot="right">
+        <div class="dropdown">
+          <button
+            class="btn btn-primary dropdown-toggle"
```
RosaRomeroGomez, line 8: I think we should be using `btn-secondary` for these settings buttons as well as for the expand panel button. In this way, we highlight buttons such as `run`, `provenance graph`, etc a bit more.

RosaRomeroGomez: image

client/src/views/Simulation/components/ModelPanel.vue
```diff
@@ -0,0 +1,122 @@
...
+
+  const components = {
+    Counters,
+    GlobalGraph,
+    SettingsBar,
+  };
+
+  // List of available graphs in a model
+  enum VIEWS {
+    CAUSAL = 'causal',
```
RosaRomeroGomez, line 61: This will change to FN, Petri Net Classic, but we can leave it like this for now.

client/src/views/Simulation/components/RunButton.vue
```diff
@@ -0,0 +1,86 @@
...
+      class="btn btn-primary"
+      :class="{ 'blue' : !autoRunInput }"
+      @click="onClickRun"
+    >
+      <font-awesome-icon :icon="['fas', ( autoRunInput ? 'pause' : 'play') ]" />
+      {{ displayText }}
+    </button>
+    <button
+      type="button"
+      class="btn btn-primary blue dropdown-toggle dropdown-toggle-split"
```
RosaRomeroGomez, line 14: I think we need to make the caret icon bigger.

client/src/views/Simulation/components/RunButton.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+  <div class="btn-group">
+    <button
+      type="button"
+      class="btn btn-primary"
+      :class="{ 'blue' : !autoRunInput }"
+      @click="onClickRun"
+    >
+      <font-awesome-icon :icon="['fas', ( autoRunInput ? 'pause' : 'play') ]" />
```
RosaRomeroGomez, line 9: I like this idea, quite nice, however, we should probably preserve the same blue when the button is in auto-run mode. Currently, it switches to gray so I'd do : `class="btn btn-primary blue"`

client/src/views/Simulation/components/RunButton.vue
```diff
@@ -0,0 +1,86 @@
...
+    />
+    <div class="dropdown-menu dropdown-menu-right" :class="{ 'show': dropdownOpen }">
+      <input type="checkbox" v-model="autoRunInput" id="auto-run-input" />
+      <label
+        for="auto-run-input"
+        title="Enable model to run automatically on when Parameters are updated">
+        Auto-Run
+      </label>
+      <div class="dropdown-divider" />
+      <h6 class="dropdown-header">_Sequence_ Settings</h6>
```
RosaRomeroGomez, line 25: `Model Execution Settings` instead I'd say.

client/src/views/Simulation/components/RunButton.vue
```diff
@@ -0,0 +1,86 @@
...
+      <div class="run-config">
+        <label for="run-config-start">Start</label>
+        <input type="number" id="run-config-start" v-model.number="configInput.start" />
+        <label for="run-config-end">End</label>
```
RosaRomeroGomez, line 29: It doesn't seem changing a value here is getting reflected in the variables time-series charts. How is this getting passed to the variables panel?

RosaRomeroGomez (commented):
I think that instead of showing the `variables` panel empty by default, we can show a message to the user informing that to get variables results, they need to click on the `run` button. Someething like:

 <div class="alert alert-info" role="alert">
      Click `Run` to get variables output. 
    </div>

client/src/views/Simulation/components/RunButton.vue
YohannParis: You need to re-run the simulation are those are settings on the run button.
It will be easy to do the re-run on value changes.

client/src/views/Simulation/components/RunButton.vue
YohannParis: Perfect. That's why I added the `_` to be sure to found a better name.

client/src/views/Simulation/components/ModelPanel.vue
YohannParis: I did not do that because in the future GroMet will support many different format, and that design will not be quickly updated when adding those formats.

YohannParis:
> I think that instead of showing the `variables` panel empty by default, we can show a message to the user informing that to get variables results, they need to click on the `run` button. Someething like:
> 
> ```
> <div class="alert alert-info" role="alert">
>      Click `Run` to get variables output. 
>    </div>
> ```

This is done as well!

YohannParis:
Screen Shot 2021-06-17 at 14 48 58

RosaRomeroGomez:
What about `<i class="fas fa-undo"></i>` for the `reset` icon?

client/src/store/modules/simulation.ts
```diff
@@ -45,6 +47,10 @@ const getters: GetterTree<any, HMI.SimulationParameter[]> = {
   getSimVariables (state): HMI.SimulationVariable[] {
     return state.variables;
   },
+
+  hasVariablesRuns (state): boolean {
```
mj3cheun, line 51: hmm im not a massive fan of this, i think the default (empty) state of variables could use some work to make this long list of conditionals unnecessary

i think ideally we should change the default state so that `state.variables?.[0]?.values.length` is all thats required

i could try to do this on merge; this is more a note to self than anything else

YohannParis: Exactly, I thought this could be done on a cleanup or update of the store later on, instead of fixing that up with 3 PR in progress on the simulation view.

YohannParis:
> What about `<i class="fas fa-undo"></i>` for the `reset` icon?
Sounds good!

client/src/views/Simulation/Simulation.vue
```diff
@@ -193,12 +184,46 @@
       }));
       this.setSimParameters({ parameters, count: 1 });
     }
+
+    async fetchVariables (): Promise<void> {
```
mj3cheun, line 188: maybe this should be renamed to `initializeX`, all 3 of them here. note to self

mj3cheun (approved):
A few notes but I think we can deal with it on cleanup later

client/src/views/Simulation/Simulation.vue
YohannParis: If you prefer, I use the word `fetch` because we are getting them from the Donu API.

RosaRomeroGomez (approved):
Thanks, looks good!

PR #226, opened 2021-06-16
Run aggregation

mj3cheun:
Screen Shot 2021-06-16 at 4 49 15 PM

- Refactor of simulation store (cleanup such as addition of missing types)
- Addition of result aggregation (using mean)

### DIFF
--- a/client/src/components/widgets/charts/MultiLinePlot.vue
+++ b/client/src/components/widgets/charts/MultiLinePlot.vue
@@ -8,6 +8,7 @@
 </template>
 
 <script lang="ts">
+  import { merge, cloneDeep } from 'lodash';
   import * as d3 from 'd3';
 
   import Component from 'vue-class-component';
@@ -31,8 +32,6 @@
   const DEFAULT_STYLE = {
     node: {
       fill: Colors.NODES.DEFAULT,
-      fillOther: Colors.NODES.OTHER,
-      fillAggregate: Colors.NODES.AGGREGATE,
       stroke: Colors.STROKE,
       strokeWidth: 1,
       borderRadius: 5,
@@ -41,18 +40,43 @@
       fill: 'none',
       strokeWidth: 2.5,
       strokeColor: Colors.NODES.DEFAULT,
-      strokeColorOther: Colors.NODES.OTHER,
-      strokeColorAggregate: Colors.NODES.AGGREGATE,
     },
     label: {
       text: Colors.LABELS.LIGHT,
     },
   };
 
+  const AGGREGATE_STYLE = {
+    node: {
+      fill: Colors.NODES.AGGREGATE,
+    },
+    edge: {
+      strokeColor: Colors.NODES.AGGREGATE,
+    },
+  };
+
+  const OTHER_STYLE = {
+    node: {
+      fill: Colors.NODES.OTHER,
+    },
+    edge: {
+      strokeColor: Colors.NODES.OTHER,
+    },
+  };
+
+  const mergeStyle = modifyingStyle => {
+    if (modifyingStyle) {
+      return merge(cloneDeep(DEFAULT_STYLE), modifyingStyle);
+    }
+
+    return DEFAULT_STYLE;
+  };
+
 @Component
   export default class VariablePanel extends Vue {
     @Prop({ default: 'Title' }) title: string;
     @Prop({ default: () => [[]] }) data: any;
+    @Prop({ default: () => [[]] }) dataAggregate: any;
 
     width: number;
     height: number;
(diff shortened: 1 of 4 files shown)


client/src/components/widgets/charts/MultiLinePlot.vue
```diff
@@ -41,18 +40,43 @@
...
+    }
+
+    return DEFAULT_STYLE;
+  };
+
 @Component
   export default class VariablePanel extends Vue {
     @Prop({ default: 'Title' }) title: string;
     @Prop({ default: () => [[]] }) data: any;
+    @Prop({ default: () => [[]] }) dataAggregate: any;
```
YohannParis, line 79: Why not adding the style inside the data object? to avoid different `data` prop in the future.
The `data` type can include a metadata property including `name`, `style`, etc.

client/src/store/modules/simulation.ts
```diff
@@ -94,10 +97,13 @@ const actions: ActionTree<any, HMI.SimulationParameter[]> = {
...
 
-  async getModelResults ({ getters }, model): Promise<unknown> {
-    return await Promise.all(
+  async fetchModelResults ({ state, getters, dispatch }, { model, aggregator }): Promise<void> {
+    const modelResults: SimulationResponse[] = await Promise.all(
       getters.getSimParameterArray.map(simParamArr => getModelResult(model, simParamArr)),
     );
+    dispatch('setSimVariables', donuSimulateToD3(modelResults as SimulationResponse[]));
+
+    state.variablesAggregate = donuSimulateToD3([aggregateModelResults(modelResults, aggregator)]);
```
YohannParis, line 106: Why `variablesAggregate` do not have a `dispatch()` like `variables`?

client/src/utils/DonuUtil.ts
```diff
@@ -15,9 +15,48 @@ export function donuToModel (donuModels: Donu.ModelDefinition[]): ModelInterface
...
+    aggregateValues[dataKeyName] = aggregateValues[dataKeyName].map(dataColumn => aggregator(dataColumn));
+  }
+
+  return {
+    times: modelResults[0].times,
+    values: aggregateValues,
+  };
+};
+
+export const donuSimulateToD3 = (responseArr: Donu.SimulationResponse[]): SimulationVariable[] => {
```
YohannParis, line 59: This function could be rename `donuSimulateToVariable` to match the return type

client/src/views/Simulation/components/VariablesPanel.vue
```diff
@@ -53,9 +54,7 @@
   import Counters from '@/components/Counters.vue';
   import MultiLinePlot from '@/components/widgets/charts/MultiLinePlot.vue';
 
-  import { donuSimulateToD3 } from '@/utils/DonuUtil';
-
-  import * as HMI from '@/types/types';
```
YohannParis, line 58: should we stop using this style of import in the future?

YohannParis (approved):

client/src/store/modules/simulation.ts
mj3cheun: there is no `Action` for setting the `variablesAggregate` property since I did not intend the `variablesAggregate` to be set independently of the fetchModelResults. therefore `dispatch()` is not necessary here

client/src/views/Simulation/components/VariablesPanel.vue
mj3cheun: hmmm good question, i think i made this change during a merge and merged the old style. ive got nothing against this style or the specific style of import personally.

YohannParis: _Parfait_, because I prefer the `* as HMI` way, it makes it easier to understand what we are doing in the code, and to see where the type is coming from.
Specially that with tree-shaking there is no memory issue now to do so.

client/src/store/modules/simulation.ts
YohannParis: I see, we can keep a loose interpretation of the [Action, Getters, Mutations](https://vuex.vuejs.org/vuex.png).
We should just make sure to not get to strict down the line.

client/src/components/widgets/charts/MultiLinePlot.vue
mj3cheun: i like this idea

client/src/utils/DonuUtil.ts
```diff
@@ -15,9 +15,48 @@ export function donuToModel (donuModels: Donu.ModelDefinition[]): ModelInterface
       type: ModelInterfaceType.computational,
     } as ModelInterface;
   });
-}
+};
 
-export function donuSimulateToD3 (responseArr: Donu.SimulationResponse[]): SimulationVariable[] {
+export const aggregateMean = (dataColumn: number[]): number =>
```
RosaRomeroGomez, line 20: I think we might want to use this kind of function in other places so we might want to take it to some `DataUtil`?

RosaRomeroGomez:
Some high-level comments. They can be addressed in a subsequent PR:

- I think we should show fewer labels on the `y` axis for each multi-line plot, it looks too crowded. 
- I tried around 5 different executions and the mean line gets buried behind the other lines. I think we can either force it to draw it last or set a threshold of 5 executions to start drawing only dots for executions instead of lines. 
image

client/src/utils/DonuUtil.ts
YohannParis: Why not just leverage d3 as it is loaded in the app anyway?
https://observablehq.com/@d3/d3-mean-d3-median-and-friends

client/src/components/widgets/charts/MultiLinePlot.vue
YohannParis: I see that you created a new prop instead of extending the data prop. That works as well.

client/src/utils/DonuUtil.ts
RosaRomeroGomez: Or `lodash`, that's a good point.

Issue #99, opened 2021-02-01
Start using native CSS custom properties for style customization


RosaRomeroGomez:
https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties

PR #229, opened 2021-06-18
Multiple runs on simulation parameters panel

YohannParis:
**What**
- Display multiple runs on the parameters panel.
- Starting to use CSS custom properties for design #99

**What**
- Clean up some _eslint_ warning message about VueX decorator library
- Started a new file to set colours as CSS custom properties
- Renamed `parametersMaxCount` to `numberOfSavedRuns` in the VueX store for clarity
- Removed the bottom axis for parameters to match the design sketches
- Display the current run following the same design on the variables panel

**Video**

test


client/src/views/Simulation/components/ParametersPanel.vue
```diff
@@ -184,7 +174,10 @@
       const yMinMax = [marginY, this.graphHeight() - marginY];
 
       // X & Y Scales
-      const xScale = param => d3.scaleLinear(d3.extent(this.runs, d => d[param]), xMinMax);
+      const xScale = param => {
+        const minMax = d3.extent(runs, d => d[param]) as Iterable<d3.NumberValue>;
+        return d3.scaleLinear(minMax, xMinMax);
```
RosaRomeroGomez, line 179: Maybe you want to use `axis` here from `utils/SVGUtil`

RosaRomeroGomez:
A couple of comments:

- I missed that in earlier PRs but you should be able to leverage `createChart` from `utils/SVGUtil` as well within `ParametersPanel` ( in `drawGraph`)
- This is just a thought to make the ParametersPanel code more reusable in the future but not mandatory to do now. We might want to take out the `svg` part of the `ParametersPanel` and create a component `ParallelCoordinates` under `components/widgets/charts/` so if we ever want to use the parallel coords viz anywhere else we can just plug it.  This is how we are doing it in the `VariablesPanel` currently.

RosaRomeroGomez (commented):
This is looking good but I had some expectations for the Params panel:
- [x] When you change a parameter value, that change should get reflected automatically in the parallel coords viz (even if I don't want to save that configuration) --> this is not included
- [x] Then, if you click `save`, that configuration gets saved so the next time I do a parameter change, I get a new line (in total now we have 2 lines, the saved one and the one I'm just playing with) --> this is what you already got.

mj3cheun (approved):
looks good, great cleanup job on the naming and types

client/src/views/Simulation/components/ParametersPanel.vue
YohannParis: unfortunately I can't, because `SVGUtils.axis()` uses `d3.nice()` and mess up the whole rounding up to power of 10 I do to display the parameters value in a nice way.

YohannParis:
> This is just a thought to make the ParametersPanel code more reusable in the future but not mandatory to do now. We might want to take out the `svg` part of the `ParametersPanel` and create a component `ParallelCoordinates` under `components/widgets/charts/` so if we ever want to use the parallel coords viz anywhere else we can just plug it.  This is how we are doing it in the `VariablesPanel` currently.

I can't really do that, as this piece of code is really tie to the panel. It will be combursome to refactor as a component or function. This part can be done by the next person that needs to create another _Parallel Coordinates_ graph (which most likely be me)

YohannParis:
New behaviour @RosaRomeroGomez:

https://user-images.githubusercontent.com/636801/122946112-15fd6980-d347-11eb-8de0-0b06dde32376.mov

RosaRomeroGomez (approved):

PR #231, opened 2021-06-21
[FEATURE ] Add bio path query

adamocarolli:
### Description
Add path queries to the bio view. See attached screenshots for examples.

### Changes
- Upgrade bgraph library from v0.1.0 to v0.2.0. See release notes for changes: https://github.com/uncharted-aske/bgraph/releases/tag/v0.2.0
- Add path query execution builder to bgraph utils. **See consideration.**
- Refactored edge/node/path cases into their own utility for code clarity and de-duplication.
- Handle the results of a path query different from the results of an edge/node query. In a path query the bgraph query results are flattened to include the underlying path present within the Gremlin's state. In an edge/node query the underlying Gremlin's path state is ignored.
- A path query clause is broken up into a pre- path clause and a -post path clause before running the path query builder function. This is because to create an n-hop path query in bgraph you need to first filter down the result space using your "from" value in the query and then later on in the query select which results to store based on your "to" value. 

### Considerations
- **Design consideration:** When a user selects a path query, edge/node filters are treated as filters along each step of the path; pruning potential paths that do not satisfy the filters.
- Edge/Node filters are applied to every node/edge along the path; thus it is not possible via the current interface to make queries that have at least one node/edge along the path that fits a criteria. Additionally, any filter queries that only act on certain edges/nodes along the path are not possible via the current interface.
- The paths that get outputted have a maximum length of 6 hops. This value can be increased, however, it comes with a negative performance trade-off.

### Testing
- Ensure that you run `yarn install` prior to testing as the bgraph module has been updated from 0.1.0-> 0.2.0

### Screenshots
Screen Shot 2021-06-18 at 4 17 24 PM
Screen Shot 2021-06-18 at 4 23 27 PM

### DIFF
--- a/client/src/utils/BGraphUtil.ts
+++ b/client/src/utils/BGraphUtil.ts
@@ -11,7 +11,7 @@ import { QUERY_FIELDS_MAP } from '@/utils/QueryFieldsUtil';
 import { isEmpty } from './FiltersUtil';
 import { buildHighlightClusterLayer, buildHighlightNodeLayer } from './GraferUtil';
 import * as FiltresUtil from '@/utils/FiltresUtil';
-import { BIO_EDGE_TYPE_OPTIONS } from '@/utils/ModelTypeUtil';
+import * as BGraphQueryUtil from '@/utils/BGraphQueryUtil';
 
 const deepCopy = (inObject, keyBlackList?: Array<any>): any => {
   let value, key;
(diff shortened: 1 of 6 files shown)


client/src/utils/BGraphUtil.ts
```diff
@@ -54,24 +54,31 @@ export const loadBGraphData = (nodesPath: string, edgesPath: string): Promise<an
...
-  // Edge Terms
-  [QUERY_FIELDS_MAP.BIO_EDGE_PRE.field]: 3,
-  [QUERY_FIELDS_MAP.BIO_EDGE_TESTED.field]: EDGE_PRIORITY_RANK,
-  [QUERY_FIELDS_MAP.BIO_EDGE_TYPE.field]: EDGE_PRIORITY_RANK,
-  [QUERY_FIELDS_MAP.BIO_EDGE_DOI.field]: EDGE_PRIORITY_RANK,
-  [QUERY_FIELDS_MAP.BIO_EDGE_POST.field]: 5,
+  [QUERY_FIELDS_MAP.BIO_NODE_POST.field]: 6,
+  // Path Terms
+  [QUERY_FIELDS_MAP.BIO_PATH_PRE.field]: 0,
+  [QUERY_FIELDS_MAP.BIO_PATH_PRE_EDGE_FILTER_LOOP.field]: 1,
```
mj3cheun, line 78: just double checking, you sure you want `BIO_PATH_PRE_EDGE_FILTER_LOOP` and `BIO_PATH_PRE_NODE_FILTER_LOOP`  to have the same rank as `BIO_EDGE_PRE` and `BIO_EDGE_POST` respectively? order is not guaranteed when the ranks are the same

adamocarolli: I've been thinking of the architecture such that the PRE/POST fields of each query type `node`, `edge` and `path` are not meant to be used in multiple query type builders. So this shouldn't be an issue.

client/src/utils/BGraphUtil.ts
```diff
@@ -83,38 +90,13 @@ export const executeBgraphEdges = (bgraphEdgeQuery: any, clause: Filter): any =>
...
     case QUERY_FIELDS_MAP.BIO_EDGE_TESTED.field: {
-      return bgraphEdgeQuery.filter(document => {
-        if (document._type === 'edge') {
-          const edge = document;
-          return edge.tested === Boolean(clause.values[0]);
-        }
-        // Document is not an edge
-        return false;
-      });
+      return BGraphQueryUtil.bioEdgeTested(bgraphEdgeQuery, clause);
```
mj3cheun, line 93: nice!

client/src/utils/BGraphUtil.ts
```diff
@@ -125,6 +107,56 @@ export const executeBgraphEdges = (bgraphEdgeQuery: any, clause: Filter): any =>
...
+    case QUERY_FIELDS_MAP.BIO_PATH_PRE.field: {
+      return BGraphQueryUtil.bioPathPre(bgraphPathQuery, clause);
+    }
+    case QUERY_FIELDS_MAP.BIO_PATH_PRE_NODE_FILTER_LOOP.field: {
+      return BGraphQueryUtil.bioPathPreNodeFilterLoop(bgraphPathQuery, clause);
+    }
+    case QUERY_FIELDS_MAP.BIO_PATH_PRE_EDGE_FILTER_LOOP.field: {
+      return BGraphQueryUtil.bioPathPreEdgeFilterLoop(bgraphPathQuery, clause);
+    }
+    case QUERY_FIELDS_MAP.BIO_PATH_POST.field: {
```
mj3cheun, line 124: up to you but i would choose to put this case last to mirror the order in which these cases might be executed. thats just personal preference tho.

client/src/utils/BGraphUtil.ts
```diff
@@ -220,26 +207,71 @@ export const filterToBgraph = (bgraph: any, filters: Filters): any => {
       clauses.push(QUERY_FIELDS_MAP.BIO_EDGE_POST as unknown as Filter);
     }
 
+    let hasPathFilters = false;
+    for (const clause of clauses) {
+      if (filterTermToPriorityRank[clause.field] === PATH_PRIORITY_RANK) {
```
mj3cheun, line 212: i would probably have done

```
clause.field === QUERY_FIELDS_MAP.BIO_PATH.field
```

because the priority rank is meant to be used for priority rank only and is not meant to be used as an ID. but this works too and theres no real not to do it this way so this is up to you as well.

client/src/utils/BGraphUtil.ts
mj3cheun: just making sure this was intentional, if it is then its fine

PR #232, opened 2021-06-22
Gromet integration

RosaRomeroGomez:


client/src/store/index.ts
```diff
@@ -1,23 +1,26 @@
...
+import * as query from './modules/query';
+import * as simulation from './modules/simulation';
 
 Vue.use(Vuex);
 
 const store = new Vuex.Store({
   modules: {
-    simulation,
     filtres,
+    knowledgeGraph,
```
RosaRomeroGomez, line 15: So we have a `models` store and a `knowledgeGraph` store. Maybe we can actually change the name of the `models` store to singular --> `model` to be consistent.

client/src/types/typesGraphs.ts
```diff
@@ -45,32 +45,32 @@ export interface GraphNodeDataInterface {
...
-nodes: GraphNodeInterface[],
-edges: GraphEdgeInterface[],
+  nodes: GraphNodeInterface[],
+  edges: GraphEdgeInterface[],
 }
 
 export interface ModelGraphInterface {
-abstract: GraphInterface,
-detailed: GraphInterface
+  abstract: GraphInterface,
```
RosaRomeroGomez, line 72: We are not going to need this interface anymore because now we have different modeling frameworks (Petri Nets, GrFNs, etc.)

client/src/types/typesModel.ts
```diff
@@ -0,0 +1,49 @@
+/**
+ * Define Types used for Computional Models
```
RosaRomeroGomez, line 2: `Computational`

PR #233, opened 2021-06-23
[FEATURE] Add path queries to Epi View

adamocarolli:
### Description
Add path queries to the epi view. See attached screenshots for examples.

### Changes
- Moved BGraph query builder functions into BGraphQueryUtil. 
- Remove hard-coding of subgraph queries in EPI view.
- Add SimpleSIR Gromet model to model store list
- Add SimpleSIR GrFN/PNC model json to static file server
- Retrieve SimpleSIR BGraph files from S3
- Refactor BGraph utilities to separate BIO logic/EPI logic
- Connect epi's global graph component with queries coming from bgraph
- Add epi path query logic

### Considerations
- Currently we are loading the bgraph data payloads from S3 whereas we are loading the gromet graph representation from static file server. We likely want to put these in the same place.
- We may not need to load the gromet graph representation as standalone and instead retrieve that representation from bgraph using an empty `bgraph.v().run()` query. Note: This may not work if the graph has duplicate edges that can not be differentiated via an id. 

### Testing
- Ensure that you update your env.local file as this makes changes to the ENV variables required.

### Screenshots
Screen Shot 2021-06-23 at 4 49 16 PM
Screen Shot 2021-06-23 at 4 48 43 PM


client/src/utils/BGraphUtil.ts
```diff
@@ -211,16 +80,32 @@ export const filterToBgraph = (bgraph: any, filters: Filters): any => {
...
         prePathClause.values = clause.values[0]; // Set "from" value
         clauses.push(prePathClause);
+
         // Set value on -POST path query clause
         const postPathClause = QUERY_FIELDS_MAP.BIO_PATH_POST as unknown as Filter;
         postPathClause.values = clause.values[0]; // Set "to" value
         clauses.push(postPathClause);
         clauses.push(QUERY_FIELDS_MAP.BIO_PATH_PRE_EDGE_FILTER_LOOP as unknown as Filter);
         clauses.push(QUERY_FIELDS_MAP.BIO_PATH_PRE_NODE_FILTER_LOOP as unknown as Filter);
+      } else if (clause.field === QUERY_FIELDS_MAP.PATH_QUERY.field) {
```
adamocarolli, line 95: Change `PATH_QUERY` to `EPI_PATH`

client/src/utils/QueryFieldsUtil.ts
```diff
@@ -156,6 +156,36 @@ const QUERY_FIELDS_MAP: QueryFieldMap = {
...
+    baseType: 'integer',
+    lexType: 'integer',
+  },
+  EPI_PATH_PRE_NODE_FILTER_LOOP: {
+    ..._field('epiPathPreNodeFilters', 'Path Pre Node Filters'),
+    ..._searchable('Path Pre', false),
+    baseType: 'integer',
+    lexType: 'integer',
+  },
+  PATH_QUERY: {
```
adamocarolli, line 177: Change `PATH_QUERY` to `EPI_PATH`

client/src/utils/BGraphQueryUtil.ts
```diff
@@ -148,3 +149,132 @@ export function bioNodeOutDegree (bgraphPathQuery, clause: Filter): any {
...
+    }
+    // Document type is an edge
+    return false;
+  })
+    .start()
+    .unique();
+}
+// eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/explicit-module-boundary-types
+export function epiPathPreNodeFilterLoop (bgraphPathQuery, _clause: Filter): any {
+  return bgraphPathQuery.in(); // TODO: Why are edge directions backward??
```
adamocarolli, line 169: Resolve this `TODO`, edge directions should match in bgraph and the renderer. This may be an issue when parsing gromet files into bgraph files.

client/src/utils/BGraphQueryUtil.ts
```diff
@@ -148,3 +149,132 @@ export function bioNodeOutDegree (bgraphPathQuery, clause: Filter): any {
...
+export function epiPathPost (bgraphPathQuery, clause: Filter): any {
+  return bgraphPathQuery.suspend(document => {
+    if (document._type === 'node') {
+      const node = document;
+      return clause.values[1] === node.label;
+    }
+    // Document type is an edge
+    return false;
+  })
+    .repeat(15) // TODO: Implement cycle detection in bgraph to remove bio path length limit
```
adamocarolli, line 185: Remove reference to bio path in comment

client/src/utils/BGraphQueryUtil.ts
```diff
@@ -148,3 +149,132 @@ export function bioNodeOutDegree (bgraphPathQuery, clause: Filter): any {
     return false;
   });
 }
+
+// EPI PATH QUERIES
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export function epiPathPre (bgraphPathQuery, clause: Filter): any {
```
adamocarolli, line 155: Query functions should no longer be exported as they are meant to be private to the BGraphQueryUtil

client/src/views/Models/Model.vue
```diff
@@ -396,20 +195,59 @@
     drilldownRelatedParameters: any = null;
     drilldownParameters: any = null;
 
-    isSplitView = false;
     subgraph: GraphInterface = null;
+    querygraph: any = null;
```
adamocarolli, line 199: `querygraph` should be of type `SubgraphInterface`; further `subgraph`, `querygraph` and `highlights` seem to be related. Can these be merged?

client/src/views/Models/Model.vue
```diff
@@ -426,6 +264,11 @@
       return this.selectedViewId === 'causal' ? this.selectedModel?.graph?.abstract : this.selectedModel?.graph?.detailed;
     }
 
+    get grometType (): string {
+      // TODO: Reverse these
```
adamocarolli, line 268: Remove comment. No longer applies.

client/src/utils/BGraphQueryUtil.ts
```diff
@@ -1,9 +1,10 @@
 import { Filter } from '@/types/typesLex';
 import { BIO_EDGE_TYPE_OPTIONS } from '@/utils/ModelTypeUtil';
+import { QUERY_FIELDS_MAP } from '@/utils/QueryFieldsUtil';
 
-// PATH QUERIES
+// BIO PATH QUERIES
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
-export function bioPathPre (bgraphPathQuery: any, clause: Filter): any {
+export function bioPathPre (bgraphPathQuery, clause: Filter): any {
```
RosaRomeroGomez, line 7: What the `Pre` stands for?

client/src/utils/BGraphQueryUtil.ts
```diff
@@ -148,3 +149,132 @@ export function bioNodeOutDegree (bgraphPathQuery, clause: Filter): any {
...
+  })
+    .start()
+    .unique();
+}
+// eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/explicit-module-boundary-types
+export function epiPathPreNodeFilterLoop (bgraphPathQuery, _clause: Filter): any {
+  return bgraphPathQuery.in(); // TODO: Why are edge directions backward??
+}
+// eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/explicit-module-boundary-types
+export function epiPathPreEdgeFilterLoop (bgraphPathQuery, _clause: Filter): any {
```
RosaRomeroGomez, line 172: I think we use arrow functions for our methods across the utils files so we probably want to be consistent here as well.

client/src/utils/BGraphQueryUtil.ts
```diff
@@ -148,3 +149,132 @@ export function bioNodeOutDegree (bgraphPathQuery, clause: Filter): any {
...
+    return false;
+  })
+    .repeat(15) // TODO: Implement cycle detection in bgraph to remove bio path length limit
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    .filter(_ => false)
+    .unsuspend();
+}
+
+// Add filter terms here to change the order in which they are processed in bgraph
+export const EDGE_PRIORITY_RANK = 2;
```
RosaRomeroGomez, line 192: How are we choosing these rank values?

client/src/graphs/svg/renderers/EpiRenderer.ts
```diff
@@ -182,4 +183,33 @@ export default class EpiRenderer extends SVGRenderer {
...
+    nonNeighborEdges.style('opacity', 0.1);
+
+    const neighborNodes = chart.selectAll('.node-ui').filter(d => nodes.map(node => node).includes(d.id));
+    neighborNodes.style('opacity', 1);
+
+    const neighborEdges = chart.selectAll('.edge').filter(d => _.some(edges, edge => {
+      return edge.source === d.source && edge.target === d.target;
+    }));
+    neighborEdges.style('opacity', 1);
+  }
```
YohannParis, line 214: In all those we can use `forEach` instead of `filter` to avoid going four times trought the same list and simplify the code.

i.e.
*From*
```JS
const nonNeighborNodes = chart.selectAll('.node-ui').filter(d => !nodes.map(node => node).includes(d.id));
nonNeighborNodes.style('opacity', 0.1);

const nonNeighborEdges = chart.selectAll('.edge').filter(d => !_.some(edges, edge => {
  return edge.source === d.source && edge.target === d.target;
}));
nonNeighborEdges.style('opacity', 0.1);

const neighborNodes = chart.selectAll('.node-ui').filter(d => nodes.map(node => node).includes(d.id));
neighborNodes.style('opacity', 1);

const neighborEdges = chart.selectAll('.edge').filter(d => _.some(edges, edge => {
  return edge.source === d.source && edge.target === d.target;
}));
neighborEdges.style('opacity', 1);
```
*To*
```JS
chart.selectAll('.edge').forEach(d => {
  const isNeighbour = edges.some(edge => {
    return edge.source === d.source && edge.target === d.target;
  });
  d.style('opacity', isNeighbour ? '1' : '0.1');
});

chart.selectAll('.node-ui').forEach(d => {
  const isNeighbour = nodes.map(node => node).includes(d.id);
  d.style('opacity', isNeighbour ? '1' : '0.1');
});
```

client/src/utils/BGraphQueryUtil.ts
adamocarolli: Prefix and Post is Postfix. The code organization here is to enable query building; these helper functions are used to build up a bgraph query. 

If your query is a `bio path query` you'd prefix it using this helper and then postfix it using the `bioPathPost` helper.

client/src/utils/BGraphQueryUtil.ts
adamocarolli: The ranks are used to order which query clause gets built into the query statement first. For example if you want to build a query that gets all edges with a belief score greater than 0.95 in bgraph - `bgraph.v().filter({_type: 'edge'}).filter(edge => edge.belief_score > 0.95).unique().run()`. 

Where the clause in this query is `belief score > 0.95` you'd build the query in the following order:
1. Add a prefix `bgraph = bgraph.filter({_type: 'edge'})` 
2. Add the clause `bgraph = bgraph.filter(edge => edge.belief_score > 0.95)`
3. Add a postfix `bgraph = bgraph.unique().run()`

To ensure the query gets built in this order (when the clauses are passed in a random order) you'd use the ranking to sort them.

YohannParis (approved):

client/.env
```diff
@@ -7,5 +7,7 @@ S3_BUCKET=?
 
 DONU_ENDPOINT=?
 
-S3_BGRAPH_NODES_KEY=?
-S3_BGRAPH_EDGES_KEY=?
+S3_GRAFER_MODELS=?
+
+S3_BGRAPH_BIO_MODELS=?
```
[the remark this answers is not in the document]
RosaRomeroGomez: Yes, I think this is a good catch and I agree with @YohannParis

client/src/utils/BGraphQueryUtil.ts
```diff
@@ -147,4 +148,133 @@ export function bioNodeOutDegree (bgraphPathQuery, clause: Filter): any {
...
+    .unsuspend();
+};
+
+// Add filter terms here to change the order in which they are processed in bgraph
+export const EDGE_PRIORITY_RANK = 2;
+export const NODE_PRIORITY_RANK = 4;
+export const PATH_PRIORITY_RANK = 7;
+export const filterTermToPriorityRank = {
+  // BIO Edge Terms
+  [QUERY_FIELDS_MAP.BIO_EDGE_PRE.field]: 1,
```
RosaRomeroGomez, line 197: I guess if we want to be consistent then these constants should be GRAPH (for BIO) and MODEL (for EPI)

client/src/views/Models/Model.vue
```diff
@@ -396,20 +195,59 @@
...
+    }
+
+    mounted (): void {
+      this.loadData();
+    }
+
+    executeFilters (): void {
+      if (this.bgraphInstance && this.getFilters?.clauses.length > 0) {
+        const querygraph = filterToBgraph(this.bgraphInstance, this.getFilters);
+        this.querygraph = {
```
RosaRomeroGomez, line 223: `queryGraph` instead to follow camelCase notation.

client/src/views/Models/Model.vue
adamocarolli: Let's do this while handling merge conflicts with main as I believe some of these are being modified.

client/src/utils/BGraphQueryUtil.ts
adamocarolli: I've made the change to the environment file. However, I think we should create a task for the larger naming change as it goes past the scope of what has been modified here.

client/src/views/Models/Model.vue
adamocarolli: Let's address this on the merge @mj3cheun. I have a feeling this name won't stick around.

YohannParis (approved):

client/src/store/modules/models.ts
```diff
@@ -1,317 +1,317 @@
-import _ from 'lodash';
```
YohannParis: Why this file is commented out instead of being deleted?

client/src/views/Models/Model.vue
adamocarolli: Switched this to `subgraph`

client/src/views/Models/Model.vue
```diff
@@ -108,6 +85,7 @@
   import { GraphInterface, GraphLayoutInterface, GraphNodeInterface, SubgraphInterface, GraphLayoutInterfaceType } from '@/types/typesGraphs';
   import * as Model from '@/types/typesModel';
   import { CosmosSearchInterface } from '@/types/typesCosmos';
+  import { GrometTypesEnum } from '@/types/typesGroMEt';
```
YohannParis, line 88: This is not an big issue, but I try to keep GroMET as much as possible out of the HMI.
This enum would be better place in `typesModel.ts` as `GraphType` because we also use those values there:
```JS
/* Define a graph (nodes and edges) per type of a model. */
type Graph = {
  file: string,
  type: 'PetriNetClassic' | 'FunctionNetwork',
  metadata?: Array<GraphMetadata>,
  graph: Graph.GraphInterface,
};
```

YohannParis (approved):

Issue #228, opened 2021-06-17
Allow user to select parameters and variables from the model representation


RosaRomeroGomez:
- Users might want to select certain parameters to tweak and certain variables to get predictions for. 
- This can be done in two ways: 1) from the models representation panel (leftmost panel) and 2) from the var or params panels themselves.

NOTE: We can assume that by default, we display all existing params and variables.

Design sketches for variables panel:
![image](https://user-images.githubusercontent.com/10552785/122432572-71081880-cf63-11eb-8699-6208fc7b7440.png)
Design sketches for parameters panel:
![image](https://user-images.githubusercontent.com/10552785/122432655-82e9bb80-cf63-11eb-9caa-c4984219a655.png)

mj3cheun:
Temporary selector put in place until the model representation panel is complete

RosaRomeroGomez:
This task is duplicated. Going to close it.

PR #235, opened 2021-06-24
Temporary models pane selector for variables and parameters

mj3cheun:
Temporary fix addressing #228

Adds a small section to the models pane allowing the user to change which variables and parameters are hidden/visible. This is done to demonstrate the capability until a working model renderer is put into the simulate panel and the feature can be implemented for real.

Note that because of issues with accessing donu using the new gromet models, it will not be possible to test this PR unaltered. The PR was developed using an older version of the `main` branch. A small clip was made demonstrating the functionality to be added in this PR, PM me on slack for access to the clip due to github size constraints.

Screen Shot 2021-06-24 at 4 32 36 PM

### DIFF
--- a/client/src/views/Simulation/components/ModelSelector.vue
+++ b/client/src/views/Simulation/components/ModelSelector.vue
@@ -1,0 +1,44 @@
+<template>
+  <div class="d-flex model-selector">
+    <div class="flex-grow-1">
+      <div class="button-group" v-for="(parameter) of getSimParameters" @click="clickHandler(parameter)" :key="parameter.name">
+        <input type="checkbox" class="mx-2" :checked="!parameter.hidden" />
+        <label>{{parameter.name}}</label>
+      </div>
+    </div>
+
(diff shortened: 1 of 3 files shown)


client/src/views/Simulation/components/ModelSelector.vue
```diff
@@ -0,0 +1,44 @@
...
+  import { Getter } from 'vuex-class';
+  import * as HMI from '@/types/types';
+
+  @Component
+  export default class ModelSelector extends Vue {
+    @Getter getSimParameters;
+    @Getter getSimVariables;
+
+    clickHandler (parameter: HMI.SimulationParameter | HMI.SimulationVariable): void {
+      parameter.hidden = !parameter.hidden;
```
YohannParis, line 31: This is the most hardcore use of reactivity in VueX I have seen.
But if it works, it works!

YohannParis (approved):

PR #237, opened 2021-06-28
Add runs counter to the simulation variables panel

### LINKED ISSUE #230: Include number of runs in Variables panel
image

mj3cheun:
Screen Shot 2021-06-28 at 2 42 21 PM

Fixes #230

Simple addition which adds a run counter to the variables panel. The light-mode mock up uses grey text, but since I didn't find that looked great on a dark theme, I went with making the text smaller to de-emphasize it while keeping the text colour the same as everything else. The other option is to keep the text the same size and colour as everything else which I think looks fine too. Thoughts?

Note: Again because of Donu difficulties, this will not work out-of-the-gate either if you wish to test it yourself. An easy way to rig it so that works is just to replace references to `getVariablesRunsCount` in the code with some integer.

### DIFF
--- a/client/src/views/Simulation/components/VariablesPanel.vue
+++ b/client/src/views/Simulation/components/VariablesPanel.vue
@@ -6,11 +6,16 @@
         :title="countersTitle"
         :data="countersData"
       />
-      <div slot="right">
-        <button type="button" disabled class="btn btn-secondary btn-settings">Settings</button>
-        <button type="button" class="btn btn-secondary py-0 btn-settings" @click="$emit('expand')">
-          <font-awesome-icon :icon="['fas', (expanded ? 'compress-alt' : 'expand-alt')]" />
-        </button>
+      <div class="d-flex align-items-center" slot="right">
+        <small class="mr-3 run-counter" v-if="getVariablesRunsCount">
+          {{`${getVariablesRunsCount} run${getVariablesRunsCount > 1 ? 's' : ''}`}}
+        </small>
+        <div class="btn-group">
+          <button type="button" disabled class="btn btn-secondary btn-settings">Settings</button>
+          <button type="button" class="btn btn-secondary py-0 btn-settings" @click="$emit('expand')">
+            <font-awesome-icon :icon="['fas', (expanded ? 'compress-alt' : 'expand-alt')]" />
+          </button>
+        </div>
       </div>
     </settings-bar>
 
@@ -183,6 +188,10 @@
   .btn-settings {
     height: 25px;
     line-height: 0;
+  }
+
+  .run-counter {
+    color: $text-color-light;
   }
 
   .chart {
--- a/client/src/views/Simulation/components/VariablesPanel.vue
+++ b/client/src/views/Simulation/components/VariablesPanel.vue
@@ -10,7 +10,7 @@
         <small class="mr-3 run-counter" v-if="getVariablesRunsCount">
           {{`${getVariablesRunsCount} run${getVariablesRunsCount > 1 ? 's' : ''}`}}
         </small>
-        <div class="btn-group">
+        <div>
           <button type="button" disabled class="btn btn-secondary btn-settings">Settings</button>
           <button type="button" class="btn btn-secondary py-0 btn-settings" @click="$emit('expand')">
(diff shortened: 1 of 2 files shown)


client/src/views/Simulation/components/VariablesPanel.vue
```diff
@@ -6,11 +6,16 @@
...
-      <div slot="right">
-        <button type="button" disabled class="btn btn-secondary btn-settings">Settings</button>
-        <button type="button" class="btn btn-secondary py-0 btn-settings" @click="$emit('expand')">
-          <font-awesome-icon :icon="['fas', (expanded ? 'compress-alt' : 'expand-alt')]" />
-        </button>
+      <div class="d-flex align-items-center" slot="right">
+        <small class="mr-3 run-counter" v-if="getVariablesRunsCount">
+          {{`${getVariablesRunsCount} run${getVariablesRunsCount > 1 ? 's' : ''}`}}
+        </small>
+        <div class="btn-group">
```
YohannParis, line 13: @RosaRomeroGomez asked to separate those buttons.
If it's fine to keep them grouped, we should fo that in all panels.

client/src/views/Simulation/components/VariablesPanel.vue
```diff
@@ -6,11 +6,16 @@
         :title="countersTitle"
         :data="countersData"
       />
-      <div slot="right">
-        <button type="button" disabled class="btn btn-secondary btn-settings">Settings</button>
-        <button type="button" class="btn btn-secondary py-0 btn-settings" @click="$emit('expand')">
-          <font-awesome-icon :icon="['fas', (expanded ? 'compress-alt' : 'expand-alt')]" />
-        </button>
+      <div class="d-flex align-items-center" slot="right">
+        <small class="mr-3 run-counter" v-if="getVariablesRunsCount">
```
YohannParis, line 10: because you created a classname for it `.run-counter` why not add all those specification in the CSS

```HTML
<span class="run-counter" v-if="getVariablesRunsCount">
```
```CSS
.run-counter {
  color: $text-color-light;
  font-size: .9em;
  margin-right: 1em; /* Whatever mr-3 is equal to */
}

YohannParis (approved):

client/src/views/Simulation/components/VariablesPanel.vue
mj3cheun: whoops, i connected those buttons by mistake using the `btn-group` class

client/src/views/Simulation/components/VariablesPanel.vue
YohannParis: My thougt on this will need a couple of hours and beers.
No worries about it, that's why I approved the PR. Maybe next year we will have time to agree on front-end architecture and styling.

RosaRomeroGomez: Yes, we had a conversation about this long time ago and since @mj3cheun was the only FE engineer at that time, we went with this styling. I think I personally prefer to have things separate and declared at the bottom of the page for each component to keep the template as clean as possible

RosaRomeroGomez:
I like this approach of making the size of the label smaller, nice job.

On a side note, I noticed that we are using `-` to inform the user there are no available variables at the moment but we use `0` to do that in the parameters panel. I think we need to be consistent and use either 0 or `-`. I think I might prefer the `-` since I think better conveys the idea of the data not being available at the moment. Any thoughts on this @YohannParis ?

RosaRomeroGomez (approved):
Thanks!

YohannParis:
> I like this approach of making the size of the label smaller, nice job.
> 
> On a side note, I noticed that we are using `-` to inform the user there are no available variables at the moment but we use `0` to do that in the parameters panel. I think we need to be consistent and use either 0 or `-`. I think I might prefer the `-` since I think better conveys the idea of the data not being available at the moment. Any thoughts on this @YohannParis ?

I agree, I implemented this solution on the variables, and forgot to do it on the Parameters as well, because we never had a situation where they were missing, but not this needs to be fixed.

mj3cheun:
cool thanks everyone, merging

PR #238, opened 2021-06-28
GroMET metadata

YohannParis:
**What**
- Display GroMET metadata in the metadata panel of models

**How**
- clean up the typesGromet
- created a date time utility to read GroMET weird use case of ISO 8601 extended format
- Keep consistent design from the rest of the vies as much as possible

**Screenshot**

test


client/src/static/mockedDataDemo.ts
```diff
@@ -11,6 +11,8 @@ export const staticFileURLs = [
 
 // eslint-disable-next-line
 export const buildInitialModelsList = ({ SIR_PN, SIR_FN }: any): Model.Model[] => {
+  const PN = GroMEt2Graph.parseGromet(SIR_PN);
+  const FN = GroMEt2Graph.parseGromet(SIR_FN);
```
mj3cheun, line 15: @adamocarolli feel free to chime in here, i believe the pre-parsed gromets are available on minio/d3?

adamocarolli: They are not, but they can be uploaded by running this script in an offline process and then saving the files to minio. This would avoid the need to do these transforms in the client and would reduce the network cost of transferring the files. 

@YohannParis if you plan to go this route let me know and we can discuss. I think the change would take about an hour or two.

YohannParis: To be honest this a refactor of @RosaRomeroGomez implementation.
I'll prefer to move it to Minio, as well as the parsing into the upload of GroMET process.
But I guess, it that for rapid prototyping and while GroMET is still in development it might be overkill. I let Rosa decide.

client/src/static/mockedDataDemo.ts
```diff
@@ -11,6 +11,8 @@ export const staticFileURLs = [
 
 // eslint-disable-next-line
 export const buildInitialModelsList = ({ SIR_PN, SIR_FN }: any): Model.Model[] => {
+  const PN = GroMEt2Graph.parseGromet(SIR_PN);
```
RosaRomeroGomez, line 14: These names won't scale if we add more models, right? We will have different `PN` and `FN` for several models so we will need more specific names.

client/src/utils/DateTimeUtil.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * Transforms a Date into a string to be displayed in the following format:
+ * Month DD, YYYY HH:MM (GMT±zz)
+ * @param date {string | Date}
+ * @returns {Date}
+ */
+export const nicePrintableDate = (datetime: string | Date): string => {
```
RosaRomeroGomez, line 7: Maybe we can find a more descriptive name? Just `formatDate`?

client/src/views/Models/components/MetadataPanel.vue
```diff
@@ -1,32 +1,53 @@
...
-       <div class="key">Knowledge</div>
-      <div><a class="text-link" :href="metadata.knowledge">{{metadata.knowledge}}</a> </div>
-    </div> -->
+  <div class="metadata-panel">
+    <details v-for="(datum, index) in metadata" :key="index">
+
+      <template v-if="isTypeModel(datum)">
+        <summary :title="datum.uid">Model</summary>
+        <h6>Variables</h6>
+        <p>{{ datum.variables.join(', ') }}</p>
```
RosaRomeroGomez, line 8: Maybe we can move this to a computed property or create a filter for it?

client/src/views/Models/components/MetadataPanel.vue
```diff
@@ -1,32 +1,53 @@
...
+        <h6>Parameters</h6>
+        <p>{{ datum.parameters.join(', ') }}</p>
```
RosaRomeroGomez, line 10: Same here.

client/src/views/Models/components/MetadataPanel.vue
```diff
@@ -1,32 +1,53 @@
...
+        <h6>Initial Conditions</h6>
+        <p>{{ datum.initial_conditions.join(', ') }}</p>
```
RosaRomeroGomez, line 12: And here.

client/src/views/Models/components/MetadataPanel.vue
```diff
@@ -1,32 +1,53 @@
...
-    <div class="metadata-item">
-       <div class="key">Knowledge</div>
-      <div><a class="text-link" :href="metadata.knowledge">{{metadata.knowledge}}</a> </div>
-    </div> -->
+  <div class="metadata-panel">
+    <details v-for="(datum, index) in metadata" :key="index">
+
+      <template v-if="isTypeModel(datum)">
+        <summary :title="datum.uid">Model</summary>
+        <h6>Variables</h6>
```
RosaRomeroGomez, line 7: I'm not a big fan of underlined text and to be consistent. What about just using `font-weight: bold` instead for these `h6` titles?

client/src/views/Models/components/MetadataPanel.vue
```diff
@@ -1,32 +1,53 @@
...
+        <summary :title="datum.uid">Code</summary>
+        <h6>Reference</h6>
+        <p>{{ datum.global_reference_id.type }} {{ datum.global_reference_id.id }}</p>
+        <h6>Files</h6>
+        <ol>
+          <li
+            v-for="(file, index) in datum.file_ids"
+            :key="index"
+            :title="file.uid"
+          >{{ file.name }} ({{ file.path }})</li>
```
RosaRomeroGomez, line 25: I need to ask where this file lives at the moment because it would be nice to be able to click on it.

client/src/views/Models/components/MetadataPanel.vue
```diff
@@ -1,32 +1,53 @@
...
+        </ol>
+      </template>
+
+      <template v-else-if="isTypeDocuments(datum)">
+        <summary :title="datum.uid">Documents</summary>
+        <details v-for="(doc, index) in datum.documents" :key="index">
+          <summary :title="doc.uid">{{ doc.bibjson.title }}</summary>
+          <h6>Source</h6>
+          <p><a :href="doc.bibjson.website.url">{{ doc.bibjson.type }}</a></p>
```
RosaRomeroGomez, line 34: I think we should embed this url in the title and we can get rid of this source field. Something like:

`        <summary :title="doc.uid"><a :href="doc.bibjson.website.url">{{ doc.bibjson.title }}</a></summary>
`

client/src/views/Models/components/MetadataPanel.vue
```diff
@@ -1,32 +1,53 @@
...
+          <h6>File</h6>
+          <p><a :href="doc.bibjson.file_url">{{ doc.bibjson.file }}</a></p>
+          <h6>Author(s)</h6>
+          <p>{{ doc.bibjson.author.map(a => a.name).join(', ') }}</p>
+          <h6>Reference</h6>
```
RosaRomeroGomez, line 39: I'd move this reference below the title, similar to how we do it in the Knowledge space.
image

client/src/static/mockedDataDemo.ts
RosaRomeroGomez: Yes, I know we talked about moving this to minio so we wouldn't need to do the parsing on our side. However, we still don't have the final GroMEts so we can keep doing it this way until everybody has uploaded their final GroMEts to that github repo. Does that sound ok? I want to avoid the scenario where we need to re-run the script many times to get the latest version uploaded in minio. However, I can see the option where we build some pipeline to get those from the github repo, parse them and put it into minio. Up to you guys. I'd leave this PR as it is for now though.

@adamocarolli @mj3cheun @YohannParis

client/src/static/mockedDataDemo.ts
YohannParis: True, this is just for now. Once we add more models etc. we can revise the code.

client/src/views/Models/components/MetadataPanel.vue
YohannParis: I can make the text bold instead of underlined.
And I can replace the `<h6>` for a non-descriptive `<span>`.

In the future 2022 version we have to be careful about semantics and accessibility with all those `<div>` and `<span>` soup.

client/src/views/Models/components/MetadataPanel.vue
YohannParis: I completely agree!

client/src/views/Models/components/MetadataPanel.vue
YohannParis: Thank you for the screenshot, COSMOS is dead so I couldn't reference how it is done.
Maybe we should use the same component to display document.

YohannParis:
Changing done

Screen Shot 2021-06-29 at 14 18 07

YohannParis:
- [ ]  I think we should keep the collapsible panels open by default since we don't have a lot of metadata to display.

- [ ]   I found the following bug that can be reproduced following these steps: 1) Select the SIR model and switch to the FN version; 2) Open the metadata panel on the left to see the metadata corresponding to the FN version; 2) Click on `Models` in the top navigation bar and then, select again the SIR model. By default you get the PNC version of the model; 3) Open the metadata panel on the left. It seems that the metadata panel still displays the metadata for the FN version. This should be updated according to the type of GroMEt selected.

This was an issue where the toggle to select which model to view was not being set properly on mounted().

RosaRomeroGomez (approved):
Nice job! Thanks.

PR #240, opened 2021-06-30
Collapse top-level boxes by default for GroMEts

RosaRomeroGomez:
**What**
- Collapse top-level nodes for GroMEts by default (dependent on the selected layout). 
- Added Algebraic Julia GroMEts (SEIRD, SIRD, SEIR) . 
image
image

**Testing**
- Select SIR model.
- Open the GrFN version of the SIR model.
- Switch between layouts.

**Note**
There seems to be some issues with collapsing multiple nodes at the same time in the svg-flowgraph library. I had to do a hack and space the calls to the `collapse` functions to address them. Additionally, it seems that after collapsing the nodes using the `dagre` layout, you can't expand them. Something I haven't figured out how to solve just yet. I might need to address these issues in the library directly. Any suggestions or help are welcome.


client/src/graphs/svg/dagre/adapter.js
```diff
@@ -16,7 +16,7 @@ export default class DagreAdapter {
         node.height = this.nodeHeight;
 
         // Special case for parameters
-        if (node.data.role?.includes('parameter')) {
+        if (node.data && node.data.role?.includes('parameter')) {
```
YohannParis, line 19: you can chain those optional chaining:
```JS
node?.data.role?.includes('parameter')
```

client/src/graphs/svg/util.js
```diff
@@ -23,7 +24,7 @@ export const traverse = (root, callBackFn, depth = 0) => {
   callBackFn(root, depth);
   if (root.nodes) {
     const d = depth + 1;
-    for (let i = 0; i < root.nodes.length; i++) {
+    for (let i = 0; i < root.nodes?.length; i++) {
```
YohannParis, line 27: You might get an issue if `root.nodes?.length` returns `undefined`.
Why not using [forEach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach)?
```JS
root.nodes?.forEach(node => traverse(node, callBackFn, depth + 1));
```

YohannParis (approved):
Apart of some subjective code style, works for me!

client/src/graphs/svg/dagre/adapter.js
RosaRomeroGomez: Thanks.

client/src/graphs/svg/util.js
RosaRomeroGomez: Done.

YohannParis (approved):

client/src/views/Models/components/Graphs/GlobalGraph.vue
```diff
@@ -118,12 +121,34 @@
...
+            }, 500);
+          });
+        };
+
+        collapsedIds.reduce((accumulatorPromise, nextID) => {
+          return accumulatorPromise.then(() => {
+            return collapseNodeTimeOut(nextID, this.renderer);
+          });
+        }, Promise.resolve());
+      }
```
mj3cheun, line 151: this whole section can be replaced by the following:

```
      collapsedIds.map((nextId, i) => {
        setTimeout(() => this.renderer.collapse(nextId), i * 500);
      });
```

regarding the hack, i looked and it seems like there is no other way to do this. the way to fix this might be to have the `collapse` function return a promise when its safe to call collapse again. that being said, currently the `collapse` function is already returning a promise, but that promise does not seem to be helpful in knowing when calling it again is ok. could this be a bug?

mj3cheun: also, is 500 ms necessary? can this (or should this) be shortened?

YohannParis: Careful as `Array.map()` returns an array, in this case `Array.forEach()` would be more appropriate.

RosaRomeroGomez: I talked to Daniel Chang (from WM) about this and he mentioned that we might be able to remove the `async` nature of the `collapse/expand` functions. I think we can try to merge this PR now and then, explore how to change that in the original svg-flowgraph library.

RosaRomeroGomez: I tried to shorten the time to 100ms but I get an error. I'm going to keep it like that and we can see if we can remove it later.

PR #241, opened 2021-07-02
Metadata panel for model nodes

YohannParis:
**What**
 - Display the metadata provided by GroMET on a _Computational Model_.

**How** 
- Created a new Vue Filter to capitalize text;
- Separated the metadata styling out of the models' components into a separate stylesheet;
- Created a new GroMET interface to handle `CodeSpanReference` metadata type;
- Tidy up the _Models_ view metadata panes to be identical; and
- Clean unused code.

**Testing**
 - Select and open a _Computational Model_ 
 - Click on a node, either in _Petri Net Classic_ or _Functional Network_ modelling framework
 - Make sure the `drilldown-panel` open on the right with the `metadata` pane open
 - See if the *Code Reference* metadata and provenance are displayed like in the screenshot.
 - If there is no metadata available, a message should be displayed

**Screenshot**

Screen Shot 2021-07-02 at 15 33 39

**Anything else**
1. Right now we get minimum information from GroMET.

```JSON
    "metadata": [
        {
          "metadata_type": "CodeSpanReference",
          "uid": "code_s_in",
          "provenance": {
            "metadata_type": "Provenance",
            "method": "Manual_claytonm@az",
            "timestamp": "2021-06-18T05:44:54:405667_MST-0700"
          },
          "code_type": "IDENTIFIER",
          "file_id": "simple_sir_code",
          "line_begin": 31,
          "line_end": null,
          "col_begin": 9,
          "col_end": null
        }
      ],
```
2. Once we have access to the code source files, we will display snippets of the code.

### DIFF
--- a/client/src/views/Models/components/DrilldownPanel/MetadataPane.vue
+++ b/client/src/views/Models/components/DrilldownPanel/MetadataPane.vue
@@ -53,10 +53,7 @@
   export default class MetadataPane extends Vue {
     @Prop({ default: null }) data: any[];
 
-    /** Clean the metadata to match our expected format.
-     * TODO: This is a temporary solution and should be replaced with
-     * a proper formatter when DONU is fully implemented.
-     */
+    /** Clean the metadata to match our expected format. */
     get metadata () : GroMET.CodeSpanReference[] {
       if (!this.data) return [];
       const cleanMetadata = this.data.map(datum => {
--- a/client/src/views/Models/components/DrilldownPanel/MetadataPane.vue
+++ b/client/src/views/Models/components/DrilldownPanel/MetadataPane.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="metadata-list">
-    <div v-if="isEmptyMetadata" class="alert alert-info" role="alert">
+    <div v-if="isEmptyMetadata" class="alert alert-info">
       No metadata at the moment
     </div>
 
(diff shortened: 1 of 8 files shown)


client/src/views/Models/components/DrilldownPanel/MetadataPane.vue
```diff
@@ -1,144 +1,110 @@
...
+  @Component
   export default class MetadataPane extends Vue {
-    @Prop({ default: null }) data: any;
-
-    showModal: boolean = false;
-    dataLoading = false;
+    @Prop({ default: null }) data: any[];
+
+    /** Clean the metadata to match our expected format.
+     * TODO: This is a temporary solution and should be replaced with
```
RosaRomeroGomez, line 57: Not sure what you mean by this comment. I don't think we will be getting metadata for nodes from donu, we will still be reading metadata from GroMEts.

RosaRomeroGomez:
This is for a subsequent PR but we need to add an outline to the selected node. I'll create a ticket for it.

client/src/views/Models/components/DrilldownPanel/MetadataPane.vue
YohannParis: From my understanding, we are using GroMET files in the front-end because DONU API will not be finished before the August demo. Isn't the final goal of HMI to not be aware of GroMET at all?
@adamocarolli, is what I'm saying accurate, or I'm just out in the boonies?

RosaRomeroGomez: You're right, we are not working GroMEts directly, however, the metadata embedded in them I don't think it will change that much.

RosaRomeroGomez (approved):

PR #259, opened 2021-07-09
Graph Evidence modal link back to Knowledge grap

### LINKED ISSUE #252: Addition of the link back to the Knowledge space from the doc modal dialog
The link can be added at the bottom of this modal dialog.
image

This is how we did it in the Models space. Instead of `Explore All`, I'd replace the link text by `Explore all in the Knowledge space`

YohannParis:
**What**
- fix #252 

**How**
- Duplicate the code of the `ModalDocMetadata` component.
- Clean up the code a bit
- fix the issue where the link to the knowledge space was broken due to the split of `docCards` and `docsClusters`
- Update both components to say _Explore all in the Knowledge space_ instead of _Explore all_

**Video**

https://user-images.githubusercontent.com/636801/125124314-e16b0b00-e0c5-11eb-9867-96038080403b.mov

**Anything else**
- I think the `ModalDocMetadata` has no reason to be a different component than the `ModalDocument`.


client/src/components/Modals/ModalDocument.vue
```diff
@@ -26,6 +26,16 @@
       </div>
       <similar-docs :doi="doi" class="m-3"/>
     </div>
+    <button
+      v-if="linkToKnowledgeSpace"
+      slot="footer"
+      type="button"
+      class="btn btn-link"
+      @click="openKnowledgeView()"
```
RosaRomeroGomez, line 34: It would be good to pass the `doi` to the `docsCards` view so we can add an automatic query there for that paper. However, I don't think we have the DOI pill in that space.

YohannParis: That make sense, I thought of that, but there is no easy way to connect the route to LexBar.
I added the DOI to the Route query nonetheless.

https://user-images.githubusercontent.com/636801/125306460-3cd30e00-e2fd-11eb-9962-18fa66f059da.mov

RosaRomeroGomez (approved):

PR #260, opened 2021-07-12
Add neighbors to node panel of bio subgraph

### LINKED ISSUE #251: Add neighborhood suggestions in the side panel on node selection in the local view
This neighborhood relationships (top 5) should be added on node selection (within the `NodePane`). We mocked this up in February and this is how it looked like. We can have something similar.

image

mj3cheun:
Add in and out neighbour lists to the bio subgraph populated using bgraph for the currently selected node. Test by selecting different nodes in the bio subgraph.

Fixes #251

https://user-images.githubusercontent.com/14062458/125315658-55dfbd00-e305-11eb-8b17-b1e1bb00c8bb.mov

Notes: This PR changes how collapsible container and collapsible item works, by changing the CSS that fits these two components to the parent container. It is likely that this change will break existing components that use these two components. From a quick search, that appears to be only the edge panel which is currently disabled. I think we should just keep this in mind, keep an eye on it, and fix any potential problems when they appear. It will be good to double check that there are no other actively used components incorporating collapsible components though.

### DIFF
--- a/client/src/views/Graphs/components/DrilldownPanel/NodePane.vue
+++ b/client/src/views/Graphs/components/DrilldownPanel/NodePane.vue
@@ -104,7 +104,7 @@
             .unique()
             .run()
             .filter(node => !subgraph.has(node.vertex.id))
-            .map(node => `${node.vertex.name} → ${this.data.label}`);
+            .map(node => `${this.data.label} → ${node.vertex.name}`);
           this.incomingNodes = bgraph
             .v({ id: this.data.id })
             .in()
@@ -112,7 +112,7 @@
             .unique()
             .run()
             .filter(node => !subgraph.has(node.vertex.id))
-            .map(node => `${this.data.label} → ${node.vertex.name}`);
+            .map(node => `${node.vertex.name} → ${this.data.label}`);
         }
       });
     }
--- a/client/src/views/Graphs/components/DrilldownPanel/NodePane.vue
+++ b/client/src/views/Graphs/components/DrilldownPanel/NodePane.vue
@@ -19,7 +19,7 @@
       </details>
 
       <details v-if="data.in_degree" class="metadata" open>
-        <summary>Incoming ({{ data.in_degree }})</summary>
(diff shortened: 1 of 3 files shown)


YohannParis:
I did a big chunk of work on the drilldown panel that was merged into main.
It will be easier to resolve those conflict before reviewing.
Sorry, I wasn't aware you were working on the same component as I was.

YohannParis (approved):

RosaRomeroGomez:
I'm getting an error when I select a node:
image
Additionally, I see that there are many duplicates here. Not sure if this comes from bgraph or from the actual dataset. Don't worry too much about this. We can fix it in a subsequent PR. 
image

client/src/views/Graphs/components/DrilldownPanel/NodePane.vue
```diff
@@ -21,11 +21,9 @@
       <details v-if="data.in_degree" class="metadata" open>
         <summary>Incoming ({{ data.in_degree }})</summary>
         <div class="metadata-content">
-          <ul>
-            <li v-for="(targetId, index) in incoming" :key="index">
-              {{ targetId }}
-            </li>
-          </ul>
+          <div class="mb-1 p-2 rounded-lg border" v-for="(rowLabel, index) in incoming" :key="index">
```
RosaRomeroGomez, line 24: Can we also have the pointer enabled when we mouse over to indicate selection?

RosaRomeroGomez:
One last comment is about displaying relationships that are already in the subgraph. I think we should remove those as well.

client/src/views/Graphs/components/DrilldownPanel/NodePane.vue
```diff
@@ -21,11 +21,9 @@
       <details v-if="data.in_degree" class="metadata" open>
         <summary>Incoming ({{ data.in_degree }})</summary>
```
RosaRomeroGomez, line 22: I think these counters are not reflecting the actual neighbors that we get from bgraph?

RosaRomeroGomez (approved):
Other than that, it looks good!

PR #261, opened 2021-07-12
Add deployment configurations

adamocarolli:
### Description
Add configurations related to enabling a Docker Swarm deployment of the HMI 

### Changes
- Add nginx template config
- Modify service endpoints depending on whether the application is running in production or development.
- Fix small linting errors that were preventing distribution builds.
- Modify client Docker image and docker-compose file to use nginx as a proxy for some of our client-side services
- Proxy API hides secrets from being exposed in client source files/network traffick. 
- Added default environment variable values and removed unneeded .env file 

### Considerations
- There are multiple ways to intercept fetch requests before sending out. In this PR I modify the endpoints in the HMI's service files depending on whether we are in 'development' or 'production'. The other approaches that can be taken are to intercept and replace the HREFs as they are being sent out; this can be done by redefining `window.fetch` or by using service workers. I find however that intercepting within the service files provided better flexibility at the cost of distributing where we are handling prod vs. development configuration logic.

### Testing
- See Confluence's HMI Deployment page for testing.


README.md
```diff
@@ -12,18 +12,20 @@ A `.env` file containing environment variables necessary for the application's o
...
 Please duplicate the `./client/.env` file into `./client/.env.local` for local development. `.env.local` has been added to the `.gitignore` to avoid unintentional upload of the environment variables.
 
-## Local Docker Swarm Deployment
-Build docker images:
-```shell script
-docker build -t docker.uncharted.software/uncharted-aske/hmi/client -f Dockerfile.client .
-```
+## Docker Compose
+The client's docker image requires configuration via environment files. The environment files that are required
+are `./client/.env.nginx` and `./client/.env.dist`. See Confluence for default configuration parameters.
```
YohannParis, line 17: Now that the project is public, we might avoid using the term _Confluence_ and use something like `Contact Uncharted...`, same thing for the _Deployment_ instructions.

client/src/services/CosmosFetchService.ts
```diff
@@ -16,40 +16,54 @@ const COSMOS_API_KEY = process.env.COSMOS_API_KEY;
 
 /// /////////////////////////////////////////////
 
-const COSMOS_API_URL = 'https://xdd.wisc.edu/sets/xdd-covid-19/cosmos/api/search';
+const COSMOS_API_URL_PREFIX = process.env.NODE_ENV === 'development'
+  ? 'https://xdd.wisc.edu'
```
YohannParis, line 20: While we are setting up DONU and S3 endpoint in the `.env` we should add COSMOS and EMMAA as well.

YohannParis (changes requested):

client/src/services/CosmosFetchService.ts
adamocarolli: We are passing COSMOS and S3 through a proxy as they contain API keys that we do not want to expose. EMMAA and Donu are public APIs so there is no reason to call them via a proxy.

YohannParis: Yes, what I meant is just to store the URL inside a environment variable like DONU and S3.

YohannParis:
From trying to install it on my local server, this process is limited to only deploy the HMI internally in Uncharted with Minio server and NGINX proxy already setup.
This does not let anybody run their own ASKE HMI on their own.

I think the description of the PR should be clearer:
> Add configurations related to enabling a Docker Swarm deployment of the HMI

to

> Add configurations related to enabling a Docker Swarm deployment of the HMI on Uncharted internal infrastructure.

YohannParis (approved):

adamocarolli:
> From trying to install it on my local server, this process is limited to only deploy the HMI internally in Uncharted with Minio server and NGINX proxy already setup.
> This does not let anybody run their own ASKE HMI on their own.
> 
> I think the description of the PR should be clearer:
> 
> > Add configurations related to enabling a Docker Swarm deployment of the HMI
> 
> to
> 
> > Add configurations related to enabling a Docker Swarm deployment of the HMI on Uncharted internal infrastructure.

Yeah you'd have to host your own AWS S3 or Minio server. Then contact Uncharted for the data to populate or create your own data for your use case.

Issue #243, opened 2021-07-06
Set parameters panel as empty by default


RosaRomeroGomez:
- EMMAA GroMEts are quite large and they include thousands of parameters. To be able to scale the parameters panel visually, we should allow the user to select parameters from the model visualization instead.
- Design:
![image](https://user-images.githubusercontent.com/10552785/124635477-7e733d00-de55-11eb-9940-194588fcdbb4.png)

PR #262, opened 2021-07-12
Updated DONU API

YohannParis:
**What**
- To be able to work on #243 I need to implement the new DONU API

**How**
- Thanks to @adamocarolli Observable and Galois DONU documentation
- Rewrite some functions and types 
- Added some fields in the mock data
- Added VueX mutations to the simulation store, so it is easier to debug with the Vue.js Dev tool.
- Change how the _aggregate_ variables are stored, to be simpler.

**Testing**
- Try to run a simulation on any model and see if it behave as expected before Galois' update.
- Be careful that if you do not select a _Functional Network_ graph before opening a simulation, Donu API will always return the same simulation results regardless of parameters values.

### DIFF
--- a/client/src/static/mockedDataDemo.ts
+++ b/client/src/static/mockedDataDemo.ts
@@ -1,7 +1,8 @@
 import { GroMEt2Graph } from 'research/gromet/tools/parser/GroMEt2Graph';
 
-import * as Model from '@/types/typesModel';
+import * as Donu from '@/types/typesDonu';
 import * as Graphs from '@/types/typesGraphs';
+import * as Model from '@/types/typesModel';
 import { getUtil } from '@/utils/FetchUtil';
 
 export const staticFileURLs = [
@@ -31,7 +32,8 @@ export const buildInitialModelsList = ({ SIR_PN, SIR_FN, SEIR_PN, SEIRD_PN, SIRD
       },
       modelGraph: [
         {
-          file: '',
+          donuType: Donu.Type.GROMET_PNC,
+          model: 'sir.gromet',
           type: Model.GraphTypes.PetriNetClassic,
           metadata: SIR_PN.metadata,
           graph: {
@@ -40,7 +42,8 @@ export const buildInitialModelsList = ({ SIR_PN, SIR_FN, SEIR_PN, SEIRD_PN, SIRD
           },
         },
         {
-          file: '',
+          donuType: Donu.Type.EASEL,
+          model: 'sir.easel',
           type: Model.GraphTypes.FunctionNetwork,
           metadata: SIR_FN.metadata,
           graph: {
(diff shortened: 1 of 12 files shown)


client/src/static/mockedDataDemo.ts
```diff
@@ -31,7 +32,8 @@ export const buildInitialModelsList = ({ SIR_PN, SIR_FN, SEIR_PN, SEIRD_PN, SIRD
       },
       modelGraph: [
         {
-          file: '',
+          donuType: Donu.Type.GROMET_PNC,
+          model: 'sir.gromet',
```
YohannParis, line 36: @RosaRomeroGomez this is what I was referring to in [Slack ](https://unchartedsoftware.slack.com/archives/C01GHFTFC4W/p1626119307034100).
I infered which model name from [Donu API `list-model`](https://observablehq.com/@adamocarolli/donu-test-1?collection=@adamocarolli/donu) response matched our mock data for the demo.

client/src/static/mockedDataDemo.ts
```diff
@@ -40,7 +42,8 @@ export const buildInitialModelsList = ({ SIR_PN, SIR_FN, SEIR_PN, SEIRD_PN, SIRD
           },
         },
         {
-          file: '',
+          donuType: Donu.Type.EASEL,
```
RosaRomeroGomez, line 45: I suppose this is temporary because we will be only executing PNCs and GrFNs for the demo, not EASEL types.

RosaRomeroGomez:
For some models such as the `SimpleSEIR` I'm getting the following error after I select it and open it in PNC form.
image

YohannParis:
> Not sure if this is relevant for this PR but for some models such as the `SimpleSEIR` I'm getting the following error after I select it and open it in PNC form.
> image

This is an issue to report to @adamocarolli, it seems that the JSONL files are wrong/corrupted.

client/src/static/mockedDataDemo.ts
YohannParis: This is based ont the model name and type from Donu `sir.easel`.

mj3cheun (approved):

YohannParis:
Added the posibility to add/remove parameters from the panel

https://user-images.githubusercontent.com/636801/125679221-42059246-eac2-40aa-bcd8-233524f0c360.mov

client/src/static/mockedDataDemo.ts
RosaRomeroGomez: Yes, but we won't be really executing `easel` type models. We will be executing `gromet-pnc` and `gromet-fn`. Currently, only `gromet-pnc` are in donu. Does this make sense to you?

RosaRomeroGomez (approved):
Nice job!

PR #264, opened 2021-07-13
Fix error when cosmos artifacts is given a doi which does not exist

mj3cheun:
to test, do an edge DOI search for 10.1016/j.intimp.2020.107018 and then click on the green edge. there should not be any error.


emilykuang (commented):
Tested on 3 different edges in the HMI, the evidence is showing up now. 
image

PR #267, opened 2021-07-14
Add screenshots for Models and Graphs lists

RosaRomeroGomez:
**What**
Just an update of the screenshots for Models and Graphs for the cards. 
I also unified the way we highlight the neighborhood for the gromets.

**Test**
Just open the `Models or `Graphs` view. 
You should get now new screenshots for cards that correspond to the new model/graph visualizations


YohannParis (approved):
I don't think it was necessary to use import, just an url would work as well.
But that's not an issue.

PR #268, opened 2021-07-14
Add/Remove simulation parameters

### LINKED ISSUE #243: Set parameters panel as empty by default

YohannParis:
**What**
- fix #243 

**How**
- Added a property `edited` to the simulation parameters to display them in the panel or not.

**Video**

https://user-images.githubusercontent.com/636801/125681968-0225cc23-ce94-4a9a-a483-2b4c178f5b18.mov


RosaRomeroGomez:
This looks good. Something I was thinking about is the ability to `Add them all/Remove them all` from the Parameters panel as well. That would mean to add two icons here:
image

YohannParis:
> This looks good. Something I was thinking about is the ability to `Add them all/Remove them all` from the Parameters panel as well. That would mean to add two icons here:
> image

Feature added.

https://user-images.githubusercontent.com/636801/125968693-050998be-5f82-4551-86a7-9c38b2acf973.mov

mj3cheun (approved):

PR #269, opened 2021-07-14
Make sure only GroMET models from Donu API are used on the HMI

YohannParis:
**What**
- filter out the non-GroMET models from the Donu APOI and the mocked data.

**How**
- use the current [Donu `model-def`](https://galoisinc.github.io/ASKE-E/donu.html#model-def) and only use `gromet-pnc` and `gromet-prt`.

### DIFF
--- a/client/src/services/DonuService.ts
+++ b/client/src/services/DonuService.ts
@@ -21,7 +21,10 @@ export const fetchDonuModels = async (): Promise<any[]> => {
   const response = await callDonu(request);
   if (response.status === Donu.ResponseStatus.success) {
     const models = response?.result as Donu.ModelDefinition[] ?? null;
-    return donuToModel(models);
+    const grometOnlyModels = models.filter(model => {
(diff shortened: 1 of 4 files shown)


client/src/services/DonuService.ts
```diff
@@ -21,7 +21,10 @@ export const fetchDonuModels = async (): Promise<any[]> => {
   const response = await callDonu(request);
   if (response.status === Donu.ResponseStatus.success) {
     const models = response?.result as Donu.ModelDefinition[] ?? null;
-    return donuToModel(models);
+    const grometOnlyModels = models.filter(model => {
+      return [Donu.Type.GROMET_PNC, Donu.Type.GROMET_PRT].includes(model.type);
```
RosaRomeroGomez, line 25: We don't want to support GROMET_PRT only `PNC` and the future `FN` SO I'd remove `Donu.Type.GROMET_PRT` from the list as well.

RosaRomeroGomez (approved):

Issue #245, opened 2021-07-06
Ability to select parameters and variables from model visualization


RosaRomeroGomez:
![image](https://user-images.githubusercontent.com/10552785/124635814-ddd14d00-de55-11eb-9914-231834a65853.png)

RosaRomeroGomez:
This has been addressed.

PR #271, opened 2021-07-15
Added double click on model representation panel to hide/show parameters and variables in simulation view

mj3cheun:
Double click nodes in the model representation panel to hide or show parameters and variables in the other panels. Solves #245

Check if PR #268 would impact this PR

Note: There is a limitation that when a node is toggled, then all parameters and variables with the label of the toggled node are toggled, which could be problematic if there is multiple parameters and variables with the same label.

https://user-images.githubusercontent.com/14062458/125829162-98332b76-c023-41a8-9bed-32f452445ccb.mov

### DIFF
--- a/client/src/views/Simulation/components/ModelPanel.vue
+++ b/client/src/views/Simulation/components/ModelPanel.vue
@@ -12,7 +12,7 @@
         </button>
       </aside>
     </settings-bar>
-    <global-graph v-if="model" :data="graph" :subgraph="highlighted" @node-dblclick="onNodeClick"/>
+    <global-graph v-if="model" :data="graph" :highlighted="highlighted" @node-dblclick="onNodeClick"/>
   </section>
 </template>
 
@@ -65,7 +65,7 @@
       ];
       return {
         nodes: this.graph.nodes.filter(node => highlightedLabels.includes(node.label)),
-        edges: this.graph.edges,
+        edges: [],
       };
     }
 
--- a/client/src/views/Simulation/components/ModelPanel.vue
+++ b/client/src/views/Simulation/components/ModelPanel.vue
@@ -12,7 +12,7 @@
         </button>
       </aside>
     </settings-bar>
-    <global-graph v-if="model" :data="graph" :highlighted="highlighted" @node-dblclick="onNodeClick"/>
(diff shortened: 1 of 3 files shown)


client/src/views/Simulation/components/ModelPanel.vue
```diff
@@ -81,6 +78,31 @@
       { name: capitalize(VIEWS.FUNCTIONAL), id: VIEWS.FUNCTIONAL },
     ];
 
+    onNodeClick (selected: Graph.GraphNodeInterface): void {
+      const selectedName = selected.label;
+      this.getSimParameters.forEach(parameter => {
+        if (parameter.metadata.name === selectedName) {
+          parameter.hidden = !parameter.hidden;
```
YohannParis, line 85: This line 85 and line 97 are where the conflict with my PR will be, instead of `hidden` it's going to be `edited`.

client/src/views/Simulation/components/ModelPanel.vue
```diff
@@ -81,6 +78,31 @@
...
+      this.getSimVariables.forEach(variable => {
+        if (variable.metadata.name === selectedName) {
+          variable.hidden = !variable.hidden;
+        }
+      });
+    }
+
+    get highlighted (): Graph.GraphInterface {
+      const highlightedLabels = [
+        ...this.getSimParameters.filter(parameter => !parameter.hidden).map(parameter => parameter.metadata.name),
```
YohannParis, line 97: Maybe you can check with the `uid` property instead of `metadata.name`?
A bit like I did here https://github.com/uncharted-aske/HMI/blob/main/client/src/store/modules/simulation.ts#L165

YohannParis (approved):

client/src/views/Simulation/components/ModelPanel.vue
mj3cheun: unfortunately this won't work. i double checked and the data being rendered by d3 (the data available here) does not contain a `uid` property. heres a sample:

Screen Shot 2021-07-19 at 1 47 27 PM

client/src/views/Simulation/components/ModelPanel.vue
mj3cheun: this will be dealt with in a seperate PR, changing to `edited` turned out to be larger task than thought.

PR #275, opened 2021-07-16
Display the search bar on the model like on the simulation view.

YohannParis:
**What**
- After changing the search bar on the simulation view, the model view was not on par.

**How**
- Clean up the Model view code and implemented the same ideas.
- Random cleanup around.

**Testing**
- The search bar is now hidden behind a toggle button, and the open simulation button match the close button on the simulation view.

### DIFF
--- a/client/src/views/Models/Model.vue
+++ b/client/src/views/Models/Model.vue
@@ -12,59 +12,98 @@
         :metadata="selectedGraphMetadata"
       />
     </left-side-panel>
-    <div class="d-flex flex-column flex-grow-1 position-relative">
-      <div class="search-row">
-        <search-bar :placeholder="`Search for model components...`"/>
-        <button class="btn btn-primary m-1" @click="onOpenSimView">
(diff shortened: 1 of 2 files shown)


mj3cheun:
I pushed a hacky fix for the searchbar which originally didnt really work (the searchbar wouldnt slide out). need more investigation to come up with an elegant fix.

YohannParis:
> I pushed a hacky fix for the searchbar which originally didnt really work (the searchbar wouldnt slide out). need more investigation to come up with an elegant fix.

Could you tell me which browser you are using. I had no issue on Firefox and Safari. I'll look into it tomorrow.

mj3cheun:
Using chrome 91

RosaRomeroGomez (approved):
Thanks!

YohannParis:
Please do not delete this branch. I would like to test and make sure why this hack was necessary.

YohannParis:
I'm submiting the fix for the hack on another PR

PR #280, opened 2021-07-20
Various fixes for the model panel

RosaRomeroGomez:
**What**
- Made counters for the variables panel consistent with the counters in the parameters panel
- Fixed issues with the highlighting of nodes in the Model panel. Double clicking adds/removes variables and parameters from the panels. 
- Commented out the tooltips until we fix them.

### DIFF
--- a/client/src/graphs/svg/renderers/EpiRenderer.ts
+++ b/client/src/graphs/svg/renderers/EpiRenderer.ts
@@ -201,8 +201,8 @@ export default class EpiRenderer extends SVGRenderer {
     chart.selectAll('.node-ui').each(function (d) {
       const isHighlighted = nodesList.map(node => node.id).includes(d.id);
       d3.select(this).select('rect')
-        .style('fill', isHighlighted ? Colors.NODES.EDITED : EpiRenderer.calcNodeColor(d))
-        // .style('stroke-width', DEFAULT_STYLE.node.strokeWidth + (isHighlighted ? 3 : 0));
+        .style('fill', isHighlighted ? Colors.NODES.EDITED : EpiRenderer.calcNodeColor(d));
+      // .style('stroke-width', DEFAULT_STYLE.node.strokeWidth + (isHighlighted ? 3 : 0));
     });
   }
 
--- a/client/src/graphs/svg/renderers/EpiRenderer.ts
+++ b/client/src/graphs/svg/renderers/EpiRenderer.ts
@@ -2,7 +2,7 @@ import * as d3 from 'd3';
 
 import { SVGRenderer } from 'svg-flowgraph';
 
-import { SVGRendererOptionsInterface, SubgraphInterface } from '@/types/typesGraphs';
+import { SVGRendererOptionsInterface, SubgraphInterface, SubgraphNodeInterface } from '@/types/typesGraphs';
 
 import { calcEdgeColor, calcLabelColor, flatten } from '@/graphs/svg/util';
 import { Colors, NodeTypes } from '@/graphs/svg/encodings';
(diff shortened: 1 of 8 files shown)


client/src/views/Simulation/components/VariablesPanel.vue
```diff
@@ -41,7 +41,7 @@
               type="button"
               @click="setSimVariableVisibility(plot.metadata.name)"
             >
-              <font-awesome-icon :icon="['fas', (plot.hidden ? 'eye' : 'eye-slash')]" />
+              <font-awesome-icon :icon="['fas', (plot.hideen ? 'eye' : 'eye-slash')]" />
```
YohannParis, line 44: typo here

client/src/graphs/svg/renderers/EpiRenderer.ts
```diff
@@ -196,15 +196,12 @@ export default class EpiRenderer extends SVGRenderer {
...
     const chart = (this as any).chart;
-    const nodes = subgraph.nodes;
-
     chart.selectAll('.node-ui').each(function (d) {
-      const isHighlighted = nodes.map(node => node.id).includes(d.id);
+      const isHighlighted = nodesList.map(node => node.id).includes(d.id);
       d3.select(this).select('rect')
-        .style('stroke', isHighlighted ? Colors.HIGHLIGHT : Colors.STROKE)
-        .style('stroke-width', DEFAULT_STYLE.node.strokeWidth + (isHighlighted ? 3 : 0));
+        .style('fill', isHighlighted ? Colors.NODES.EDITED : EpiRenderer.calcNodeColor(d));
```
mj3cheun, line 204: Did you mean to revert back to node fill highlighting? Now the simulate page looks like this.

Screen Shot 2021-07-20 at 9 20 15 AM

client/src/types/types.ts
```diff
@@ -53,6 +53,7 @@ interface SimulationParameter extends Donu.ModelParameter {
 type SimulationVariableValues = Array<{x: number, y: number}>;
 
 interface SimulationVariable extends Donu.ModelVariable {
+  edited?: boolean,
```
YohannParis, line 56: Why do we need an edited on the Variables panel?

YohannParis (approved):

mj3cheun:
Looks like if you use the hide/show button and the model representation to select/deselect nodes, its possible to make the model representation go out of sync with the parameters/variables panel

YohannParis:
I get no charts on the variables panel.
You can see on the top right I have 1 run, and I can see that Galois sends the results.

Screen Shot 2021-07-20 at 09 26 56

client/src/graphs/svg/renderers/EpiRenderer.ts
RosaRomeroGomez: image
This is how it looks for me @mj3cheun

client/src/types/types.ts
RosaRomeroGomez: We should also be able to add and remove variables from the model visualization to the Variables panel. Does this make sense @YohannParis ?

client/src/views/Simulation/components/VariablesPanel.vue
RosaRomeroGomez: Fixed

client/src/graphs/svg/renderers/EpiRenderer.ts
RosaRomeroGomez: So yes, this was intentional as discussed over slack.

client/src/types/types.ts
YohannParis: Rosa explained that now Variables behaves like Parameters and you need to double-click on the model node to display them on the Variables panel.

YohannParis:
- Added the Add/remove all on the variables panel.
- Add actions and mutations on the simulation store to handle those request.
- Hide the `Show/Hide` buttons on the parameters panel
- Update the variable button to _remove_ instead of _hide_

https://user-images.githubusercontent.com/636801/126348519-5ffb0883-7456-4253-952a-fe0c6ba54e99.mov

YohannParis (approved):

PR #282, opened 2021-07-20
Model graph selector autopopulate

### LINKED ISSUE #276: Populate modeling framework menu dynamically
- Not all models will be provided in all the formats (GrFN and PNC).
- Currently, the dropdown menu is populated manually. Instead, we want it to populate it dynamically, according to the available types of GroMET for a model. 

image

YohannParis:
**What**
- Fix #276 

**How**
- Update the Model type and Model store to use `Model.GraphTypes` instead of a number index. So now the graphs in a model do not have to be ordered. I don't think a  `Map` or `Set` is necessary and makes reactivity in Vue.js 2.0 difficult.

- Update the Simulation store and Donu Service to use the graph type to select the appropriate graph information for Donu API calls.

- On the model view, I use a const to define graph types' display names. The list is filtered based on the selected model graph type available.

- If their is no choice available: the modeling framework button is disbaled and the _caret down_ icon is removed.

**Video**

https://user-images.githubusercontent.com/636801/126505903-b2bbe718-ffbb-46ee-8190-c4d12544e5d1.mov

**Testing**
- Open a model and see if the selection of modeling framework available in the dropdown matches the one in the GroMET.

**Consideration**
- @mj3cheun, Right now, this is based on the mock data; once we pull from Donu API, we have to make sure the Model saved in VueX are proper mergers of different GroMet.

- @adamocarolli, How are we going to get BGraph information that match Donu API models?


client/src/static/gromets/SIRD_gromet_PetriNetClassic_metadata.json
```diff
@@ -195,7 +195,7 @@
     {
       "provenance": {
         "method": "FromPNC_baas@gt",
-        "timestamp": "2021-07-07T15:27:484320_UTC-0000",
+        "timestamp": "2021-07-07T15:27:48:4320_UTC-0000",
```
RosaRomeroGomez, line 198: This will then probably need to be corrected in donu side since we will be moving to ingesting the models from them?

RosaRomeroGomez (approved):
Really nice clean-up work. Thank you!

client/src/static/gromets/SIRD_gromet_PetriNetClassic_metadata.json
YohannParis: Nope, that was an issue on the GroMET file for some reason.
As far as I worked with Donu API they send the date in the proper format.
If this issue comes back, it's a simple fix.

PR #283, opened 2021-07-21
Switched from hardcoded gromets to dynamically fetched gromets from donu

mj3cheun:
Switches datasource for epi graphs from the static gromet files to dynamically fetched gromet files from donu. Also enabled fetching of fnets from donu. Thanks a lot to @adamocarolli for his help on this task.

Notes:
- I am seeing periodic CORS errors from donu while testing, keep that in mind when performing tests. For example in the recording below, i encountered a CORS error when trying to simulate
- I have not deleted the gromet static pipeline, could be useful as a backup or for testing
- It appears (correct me if im wrong) that donu does not provide model descriptions, that has been left blank for now
- Used the `any` type as a placeholder for gromet types; gromet types are available in the same place as `GroMEt2Graph` but it needs to be exported.

https://user-images.githubusercontent.com/14062458/126527207-02818e04-b2f7-4767-b6e2-66964c8b76b2.mov


client/src/services/DonuService.ts
```diff
@@ -12,19 +13,85 @@ const callDonu = (request: Donu.Request): Promise<Donu.Response> => {
...
-    const models = response?.result as Donu.ModelDefinition[] ?? null;
-    const grometOnlyModels = models.filter(model => {
-      return [Donu.Type.GROMET_PNC/*, Donu.Type.GROMET_FN */].includes(model.type);
+    // 1. Get models
+    let models = response?.result as any ?? null;
+    // 2. Filter models to PNC and FN type
+    models = models.filter(model => {
+      return [Donu.Type.GROMET_PNC, Donu.Type.GROMET_FN].includes(model.type);
     });
-    return donuToModel(grometOnlyModels);
```
YohannParis, line 27: That `donutToModel` method was supposed to encapsulate all the step you replaced it with. The goal was to separate calling Donu (with `DonuService.ts`) and manipulating Donu data (with `DonuUtil.ts`.)

No need to change it, just to let you know.

client/src/services/DonuService.ts
```diff
@@ -12,19 +13,85 @@ const callDonu = (request: Donu.Request): Promise<Donu.Response> => {
...
+        model,
+      },
+    },
+  };
+  const response = await callDonu(request);
+  if (response.status === Donu.ResponseStatus.success) {
+    const modelSource = response.result.source;
+    return JSON.parse(modelSource);
+  } else {
+    console.error('[DONU Service] — fetchDonuModelSource', response);
```
YohannParis, line 31: You need to add the following comment to avoid linting errors:
`// eslint-disable-line no-console`

client/src/types/typesDonu.ts
```diff
@@ -56,6 +57,9 @@ type ModelDefinition = {
   type: Type;
 }
 
+// TO-DO: Export GroMEt type from research/gromet/tools/types/GroMEt
```
YohannParis, line 60: Could you precise the full URL, as I cannot find /research on the codebase.

client/src/services/DonuService.ts
```diff
@@ -12,19 +13,85 @@ const callDonu = (request: Donu.Request): Promise<Donu.Response> => {
   }
 };
 
+export const getDonuModelSource = async (model: string, type: Donu.Type): Promise<Donu.ModelGraph> => {
```
YohannParis, line 16: Does this function needs to be exported?

client/src/services/DonuService.ts
```diff
@@ -12,19 +13,85 @@ const callDonu = (request: Donu.Request): Promise<Donu.Response> => {
...
+  const response = await callDonu(request);
+  if (response.status === Donu.ResponseStatus.success) {
+    const modelSource = response.result.source;
+    return JSON.parse(modelSource);
+  } else {
+    console.error('[DONU Service] — fetchDonuModelSource', response);
+  }
+};
+
+const populateDonuModelWithGrometSource = async (model: Donu.ModelDefinition): Promise<void> => {
```
YohannParis, line 35: Why this needs to be a separate `const` it looks small enough to be part of step 3 `map()`

client/src/services/DonuService.ts
```diff
@@ -12,19 +13,85 @@ const callDonu = (request: Donu.Request): Promise<Donu.Response> => {
...
     command: Donu.RequestCommand.LIST_MODELS,
   };
 
   const response = await callDonu(request);
   if (response.status === Donu.ResponseStatus.success) {
-    const models = response?.result as Donu.ModelDefinition[] ?? null;
-    const grometOnlyModels = models.filter(model => {
-      return [Donu.Type.GROMET_PNC/*, Donu.Type.GROMET_FN */].includes(model.type);
+    // 1. Get models
+    let models = response?.result as any ?? null;
```
YohannParis, line 49: You can use `Donu.ResponseResult` instead of `any` for the type of the Donu response result.

client/src/types/typesModel.ts
```diff
@@ -13,6 +13,8 @@ type Metadata = {
   id?: string,
   name: string,
   version?: string,
+  model?: string,
+  type?: string,
```
YohannParis, line 17: What are those two for? I don't see you using them in this PR.

client/src/services/DonuService.ts
```diff
@@ -12,19 +13,85 @@ const callDonu = (request: Donu.Request): Promise<Donu.Response> => {
...
+        donuType: model.type,
+        model: model.source.model,
+        type: model.gromet.type,
+        metadata: model.gromet.metadata,
+        graph: {
+          nodes: model.graph.nodes,
+          edges: model.graph.edges,
+        },
+      };
+      if (name === 'SimpleSIR' || name === 'SimpleSIR_metadata') {
```
YohannParis, line 76: Why do we need this specific behaviour? is this because Donu does not format them like the others.

client/src/services/DonuService.ts
```diff
@@ -12,19 +13,85 @@ const callDonu = (request: Donu.Request): Promise<Donu.Response> => {
...
-    return donuToModel(grometOnlyModels);
+    // 3. Populate models with model source
+    await Promise.all(models.map(model => populateDonuModelWithGrometSource(model)));
+    // 4. Transform Gromet to graph for rendering
+    models.forEach(async model => {
+      model.graph = GroMEt2Graph.parseGromet(model.gromet);
+    });
+    // TODO: 5. Transform Gromet to bgraph for querying
+    // 6. Group models by model name
+    const output = new Array(1);
```
YohannParis, line 62: What happens when Donu finally add different GroMET types for the same model?

YohannParis (approved):
I really like all the comments in the massaging of Donu response into our Model representation. ❤️

client/src/types/typesDonu.ts
mj3cheun: its not on the codebase, but it is part of the research repo

https://github.com/uncharted-aske/research/blob/main/gromet/tools/types/GroMEt.d.ts

client/src/services/DonuService.ts
mj3cheun: its called strictly within the function true, we can remove the exported and add it back if its ever needed later

client/src/services/DonuService.ts
mj3cheun: we need this because there is no way to tell which models are the same models but in different form (pnc vs fnet) using the data available on donu. this is a manual way of putting the `SimpleSIR` model together which is the only model available at the moment available in different forms.

client/src/services/DonuService.ts
mj3cheun: see comment on line 76

client/src/services/DonuService.ts
```diff
@@ -12,19 +13,85 @@ const callDonu = (request: Donu.Request): Promise<Donu.Response> => {
...
+    // 2. Filter models to PNC and FN type
+    models = models.filter(model => {
+      return [Donu.Type.GROMET_PNC, Donu.Type.GROMET_FN].includes(model.type);
     });
-    return donuToModel(grometOnlyModels);
+    // 3. Populate models with model source
+    await Promise.all(models.map(model => populateDonuModelWithGrometSource(model)));
+    // 4. Transform Gromet to graph for rendering
+    models.forEach(async model => {
+      model.graph = GroMEt2Graph.parseGromet(model.gromet);
```
RosaRomeroGomez, line 58: Here there will be an exception for emma Gromets. They are too large to just display as they are and we will need to build an aggregate version of them. Nelson has done the aggregation for those gromets so we have two options: 1) put the aggregated version of emmaa gromets in local so we can display them; 2) have a similar parser that we can use on the fly (this will probably be not that viable since Nelson would need to convert that python script to typescript and I think he doesn't know typescript). Thoughts here @adamocarolli @mj3cheun ?

client/src/services/DonuService.ts
RosaRomeroGomez: Another option would be to ask donu to fix that?

RosaRomeroGomez:
This is great. And only took you guys a day, nice work. 

One comment we will want to fix in a subsequent PR probably is about the screenshots for models. I was actually taking then manually and assigning them based on the order of models but now, that doesn't work well anymore. 
image
@mj3cheun would you mind to create a ticket about this?

client/src/services/DonuService.ts
adamocarolli: We can talk to Nelson about his code and come up with a plan for this. This shouldn't be too much of an issue.

RosaRomeroGomez: Ok, sounds good. I'll start the conversation.

client/src/types/typesDonu.ts
YohannParis: Thanks, now it will be accessible with Git blame.

client/src/services/DonuService.ts
YohannParis: I'll mark this as resolve

client/src/services/DonuService.ts
YohannParis: Why removing the type instead of using `Donu.ResponseResult`?

client/src/services/DonuService.ts
YohannParis: @RosaRomeroGomez: https://aske-e.slack.com/archives/C01D0UUU25N/p1626899002402700

PR #284, opened 2021-07-21
Add small label on the parameter axis

YohannParis:
**What**
- Display scale on the parameter panel on the simulation view.

**How**
- Save the new roundup to pow10 range of the parameters.
- When adding the axis `<line>`, we also add two `<text>` labelled with min, max of the range.
- Group both the line and the labels in a group and translate the group as a whole.
- Made the label as dark and small as possible to maintain WCAG AA.
- Use scientific notation for big numbers.

**Screenshot**

Screen Shot 2021-07-21 at 14 58 21


RosaRomeroGomez (approved):
Thanks!

PR #285, opened 2021-07-21
Move the parameter action button

YohannParis:
**What**
- Move the parameter action button from the top-right to the middle-right

**How**
- Change the CSS grid template.

**Screenshot**

Screen Shot 2021-07-21 at 16 37 30


YohannParis:
I merged this myself as it is a tiny CSS settings. Not worth wasting time.

PR #288, opened 2021-07-22
Update svg-flowgraph to the latest version and remove collapse node animation

mj3cheun:
Updated `svg-flowgraph` to latest version from 0.2.0 to 0.4.4 to take advantage of the interface change which removes render from `collapse` and `expand`, hence allowing either function to be called for a list of nodes to be collapsed/expanded without having to wait for the render for each to finish before the next can be called. This requires calling `yarn install` before running.

As this is a library update, its worth testing components using `svg-flowgraph` to ensure there aren't any bugs introduced.

Also note, there is a flash of fully expanded graph before the collapsed version is shown, I have not found a way to remove this.

https://user-images.githubusercontent.com/14062458/126690897-838eaee5-507c-4979-953f-822b6c00780c.mov

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -44,7 +44,7 @@
     "preact": "8.x",
     "research": "git+https://github.com/uncharted-aske/research.git",
     "sigma": "1.2.1",
-    "svg-flowgraph": "^0.2.0",
+    "svg-flowgraph": "^0.4.0",
     "tiny-emitter": "^2.1.0",
     "tweakpane": "^1.5.8",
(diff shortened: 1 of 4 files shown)


YohannParis (approved):
Looks good!

RosaRomeroGomez:
This is much better at least that than awkward animation at the beginning. Something we can also add in a subsequent PR is a + or something like that to the labels for boxes so it is clear that you can collapse/expand them.

RosaRomeroGomez (approved):

PR #290, opened 2021-07-25
Starting the process of multiple simulation

YohannParis:
**What**
- Display multiple models into the Simulation space.

**How**
- Use `selectedModelIds` to display multiple Models in the _Simulation_ view.
- The _Comparison_ view is broken, so instead, when multiple models are selected, a temporary _Open Simulation View_ button is displayed next to the _Comparison_ one. (c.f. the first screenshot)
- Move Store Simulation Types into the HMI type files
- Change the Simulation responsive grid to accept as many models as selected by the user. (This demands too much memory from the browser. Keep it under three models.)
- Update the Simulation store to handle multiple models.
- Update the Parameter panels to have a more natural input field.
- Disable the `Run` and `Save` button if some parameters have non numbers values.

**N.B.**
- Small fix on the EpiRendered to check a graph is available before highlighting nodes.
- Change the Parameters panel scale to `[-1, 1]` if the value is `0`.
- Added a _Loader_ on the model list and model page.
- Fix a small overflow issue on the _Global Graph_ component.
- For some reason, this is using a lot of memory, so if @mj3cheun or @adamocarolli can make sure I didn't shoot myself on the foot, that would be great.

**Screenshot**

Screen Shot 2021-07-26 at 13 54 56

**Video**

https://user-images.githubusercontent.com/636801/127382373-ed721c71-a1f0-49fa-9791-6fa3fe86d9ce.mov


client/src/views/Models/Model.vue
```diff
@@ -13,7 +13,8 @@
       />
     </left-side-panel>
 
-    <section>
+    <loader v-if="!selectedGraph" loading="true" />
```
RosaRomeroGomez, line 16: Nice, thank you.

client/src/views/Simulation/Simulation.vue
```diff
@@ -145,43 +155,68 @@
...
+        } else if (this.setSelectedModels) {
+          gridMap.push([
+            'model_' + index,
+            'model-parameters-separator',
+            'parameters_' + index,
+            'parameters-variables-separator',
+            'variables_' + index,
+          ]);
+        }
+        // gridMap.push(['row-separator']);
```
RosaRomeroGomez, line 200: Remove commented code.

RosaRomeroGomez:
This is looking good. Thanks @YohannParis. Some comments below:

- It seems we are getting the scrolling bar even when we don't have parameters added to the panel, we might want to remove it or color it as the panel background so it is less noticeable. 
- Could we also make the separation between rows for the parameters panel grid smaller? In this way, we can fit more parameters without having to scroll much. 
image

client/src/store/modules/simulation.ts
```diff
@@ -4,105 +4,118 @@ import * as Model from '@/types/typesModel';
...
-const state: SimulationState = {
+const state: HMI.SimulationState = {
   numberOfSavedRuns: 0,
-  parameters: [],
-  variables: [],
+  models: [],
+};
+
+const currentNumberOfRuns = (state: HMI.SimulationState): number => {
+  return state.models[0]?.parameters?.[0]?.values.length ?? 0;
```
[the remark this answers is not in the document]
YohannParis: This is correct, and it is a good idea to add a comment.

client/src/views/Simulation/Simulation.vue
YohannParis: This is still a draft, and there is a lot of bugs to iron out, this being one of them.

YohannParis:
> This is looking good. Thanks @YohannParis. Some comments below:
> 
>     * It seems we are getting the scrolling bar even when we don't have parameters added to the panel, we might want to remove it or color it as the panel background so it is less noticeable.
> 
>     * Could we also make the separation between rows for the parameters panel grid smaller? In this way, we can fit more parameters without having to scroll much.

Screen Shot 2021-07-27 at 16 23 44

Both have been fixed.

client/src/services/DonuService.ts
```diff
@@ -72,6 +71,7 @@ export const fetchDonuModels = async (): Promise<Model.Model[]> => {
           edges: model.graph.edges,
         },
       };
+      delete model.gromet;
```
RosaRomeroGomez, line 74: Why is this needed?

client/src/views/Simulation/components/ModelPanel.vue
```diff
@@ -39,28 +45,53 @@
...
-    @Getter getSimParameters;
-    @Getter getSimVariables;
+    @Getter getSimModel;
+    @Getter getSelectedModelGraphType;
     @Action toggleParameter;
     @Action toggleVariable;
 
     settingsOpen: boolean = false;
 
     onNodeClick (selected: Graph.GraphNodeInterface): void {
```
RosaRomeroGomez, line 55: I think this method should be probably called `onNodeDblClick` since it is not just a simple click, it gets triggered on double click.

client/src/views/Simulation/components/ModelPanel.vue
```diff
@@ -39,28 +45,53 @@
...
+            .filter(node => highlightedLabels.includes(node.label))
+            .map(node => ({ id: node.id }));
+        }
+      }
+      return [];
     }
 
     get graph (): Graph.GraphInterface {
-      return this.model?.modelGraph?.[0]?.graph;
+      // return this.model?.modelGraph?.[0]?.graph;
```
RosaRomeroGomez, line 90: Remove commented code.

RosaRomeroGomez (approved):
Great job @YohannParis

client/src/services/DonuService.ts
mj3cheun: my understanding of how garbage collection works in `js` is that when the function finishes, the `model` object will be deleted by the garbage collector which includes the `gromet` property, so this deletion should not be necessary as its done automatically. I am not 100% certain though, I can test this out tomorrow by allocating like a 3gb buffer in the `model` object and see if the memory is released after the function is finished.

client/src/views/Simulation/Simulation.vue
```diff
@@ -145,46 +159,83 @@
...
-      return this.expandedId
-      ? [[this.expandedId]]
-      : [
-          ['model', 'div1', 'parameters', 'div2', 'variables'],
-        ];
+      const gridMap: string[][] = [];
+
+      this.selectedModels.forEach(model => {
+        if (this.expandedId) {
+          gridMap.push([this.expandedId + '_' + model.id]);
```
mj3cheun, line 194: coooool programmatically generated gridmaps

client/src/services/DonuService.ts
mj3cheun: i did the buffer allocation test and i compared js heap size before and after adding `delete model.gromet`, doesnt look like this line is necessary

YohannParis: Awesome, thank you @mj3cheun

PR #300, opened 2021-07-27
Add in-memory size limit to fetch memoization system and implemented for donu service

mj3cheun:
The goal of this PR is to add memoization back to the donu simulate fetch. This will make the simulate view more responsive by eliminating most redundant API calls to donu and solves issue #297. To test, ensure the simulate page is working properly, one should notice a slight increase in responsiveness particularly with a large number of runs.

In order to address previous concerns with the current memoization system with high memory usage and memory leaks, the system was modified to track the size of each entry stored in bytes and limit the total amount of memory consumed by the store to a specified number of bytes. If the store fills up it will drop entries starting from the first one stored until there is sufficient space for a new entry to be stored.

The two parameters controlling the behaviour of the memoization store are the `storeSizeLimit` (discussed above) limiting the total size of the store; and the `storeEntryLimit` limiting the size of each entry. If an item exceeds the `storeEntryLimit` it is not stored at all.

### DIFF
--- a/client/src/utils/FetchUtil.ts
+++ b/client/src/utils/FetchUtil.ts
@@ -1,7 +1,34 @@
 import { DataFile, DataSource } from '@dekkai/data-source';
 import s3Client from '@/services/S3Service';
 
+const storeSizeLimit = 104857600; // 100MB in bytes
+const storeEntryLimit = 10485760; // 10MB in bytes
 const memoizedStore = new Map();
+const memoizedSize = new Map();
+
+const storeSize = (): number => {
+  let totalValue = 0;
+  for (const value of memoizedSize.values()) {
+    totalValue += value;
+  }
+  return totalValue;
+};
+
+const storeResult = (hash: string, result: unknown, size: number): void => {
+  if (size > storeEntryLimit) {
+    return;
+  }
+
+  // TO-DO: Optimize this
+  while (storeSize() + size > storeSizeLimit) {
+    const shiftKey = memoizedStore.keys().next().value;
+    memoizedStore.delete(shiftKey);
+    memoizedSize.delete(shiftKey);
+  }
+
+  memoizedStore.set(hash, result);
+  memoizedSize.set(hash, size);
+};
 
 export const getUtil = async (urlStr: string, paramObj?: Record<string, any>): Promise<any> => {
   const init: RequestInit = {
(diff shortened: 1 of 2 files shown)


YohannParis (approved):

client/src/utils/FetchUtil.ts
```diff
@@ -1,7 +1,34 @@
 import { DataFile, DataSource } from '@dekkai/data-source';
 import s3Client from '@/services/S3Service';
 
+const storeSizeLimit = 104857600; // 100MB in bytes
+const storeEntryLimit = 10485760; // 10MB in bytes
 const memoizedStore = new Map();
+const memoizedSize = new Map();
```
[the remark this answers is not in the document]
YohannParis: https://en.wikipedia.org/wiki/Memoization

YohannParis (approved):

client/src/utils/FetchUtil.ts
```diff
@@ -1,7 +1,34 @@
...
+};
+
+const storeResult = (hash: string, result: unknown, size: number): void => {
+  if (size > storeEntryLimit) {
+    return;
+  }
+
+  // TO-DO: Optimize this
+  while (storeSize() + size > storeSizeLimit) {
+    const shiftKey = memoizedStore.keys().next().value;
```
adamocarolli, line 24: I think we should remove the term `memoized` here. Memoization implies storing the result of a previously run function, however, this store is used mostly to store a copy of data that's origin lies on another machine.

Cache is probably the more general and well understood term.

client/src/utils/FetchUtil.ts
```diff
@@ -1,7 +1,34 @@
 import { DataFile, DataSource } from '@dekkai/data-source';
 import s3Client from '@/services/S3Service';
 
+const storeSizeLimit = 104857600; // 100MB in bytes
```
adamocarolli, line 4: Consider ALL_CAPSING these to differentiate between local and global constants.

client/src/utils/FetchUtil.ts
```diff
@@ -1,7 +1,34 @@
...
+const storeEntryLimit = 10485760; // 10MB in bytes
 const memoizedStore = new Map();
+const memoizedSize = new Map();
+
+const storeSize = (): number => {
+  let totalValue = 0;
+  for (const value of memoizedSize.values()) {
```
adamocarolli, line 11: Consider keeping track of size as items are added removed instead of recomputing each time.

client/src/utils/FetchUtil.ts
```diff
@@ -1,7 +1,34 @@
...
+  }
+  return totalValue;
+};
+
+const storeResult = (hash: string, result: unknown, size: number): void => {
+  if (size > storeEntryLimit) {
+    return;
+  }
+
+  // TO-DO: Optimize this
```
adamocarolli, line 22: Consider batching deletes so you don't need to free/defragment memory after each call.

adamocarolli:
Looks good approved!

adamocarolli (approved):

client/src/utils/FetchUtil.ts
mj3cheun: short of using `map.clear` there isnt a way of batch deleting with js maps anyway, correct me if im wrong... while loop is probably as good as it will be

PR #302, opened 2021-07-27
Use GSL engine for petri net classic simulation

adamocarolli:
### Description
Temporarily use GSL engine as default for petri net classic simulations as currently Donu API contains a bug when using Algebraic Julia as default.

### Changes
- Add simulation engine types
- Override default simulation engine for petri net classic model types

### Considerations
- This is a temporary change while Donu is being fixed.

### Testing
- I used the SimpleSIR model for testing.

### Artifacts
Screen Shot 2021-07-27 at 4 40 28 PM

### DIFF
--- a/client/src/services/DonuService.ts
+++ b/client/src/services/DonuService.ts
@@ -146,6 +146,14 @@ export const getModelVariables = async (model: Model.Model, selectedModelGraphTy
   }
 };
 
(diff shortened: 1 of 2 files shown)


YohannParis (approved):
Great Job !

RosaRomeroGomez (approved):
Thanks!

PR #309, opened 2021-07-28
[FEATURE] Add querying to each model

adamocarolli:
### Description
Initialize bgraph instance for each model to enable querying.

### Changes
- Transform Gromet files into bgraph format.
- Initialize bgraph instance using bgraph data input.

### Considerations
- Alternatively we can do Gromet file transformations in an offline process. However, I believe this to be a good intermediate solution.

### Testing
- No special instructions.

### Artifacts
Screen Shot 2021-07-28 at 3 29 10 PM
https://user-images.githubusercontent.com/15199528/127389479-1196ba50-d6a6-435b-b859-801ed504d029.mov


RosaRomeroGomez:
Logging a comment here so we don't forget:
- We probably want to add a message (via a toaster or some alert) to inform the user when no paths are found.

PR #315, opened 2021-07-29
Display other types of GroMEt metadata types

### LINKED ISSUE #306: Display other types of metadata for GroMEts
Currently, it seems we only include the metadata type `CodeSpanReference`. However, GrFN GroMets include other types that have some interesting pieces of information such as `EquationDefinition`, `TextDefinition`, and `TextualDocumentReferenceSet` (this is not an exhaustive list but it should include the most important ones).

Particularly, `TextualDocumentReferenceSet` should include `aske_id` that we can use to query xDD and retrieve relevant information to display in the `Knowledge` tab (KnowledgePane).  

NOTE: Only GrFNs include this type of metadata types so you can only work with SIR GrFN GroMet at the moment.

YohannParis:
**What**
- fix #306 
- This is mostly for _Functional Network_ GroMEts

**How**
- Add other GroMEt MetadataType `EquationDefinition`, `TextDefinition`, and `TextParameter`.
- If there is no metada, the drilldown does not open on the _metadata_ tab.
- Display those new types in the metadata drilldown.
- The Metadata are displayed in the followind order based on availability: 
  `TextDefinition` > `TextParameter` > `EquationDefinition` > `CodeSpanReference`

**Screenshot**
- A box on the SIR model
Screen Shot 2021-07-30 at 12 00 40

- The variable Beta
Screen Shot 2021-07-30 at 12 00 56


RosaRomeroGomez:
I've noticed two things and one of them might be just my fault? 
- For the SIR model in PNC format, since we don't have metadata, when I click on the SIR box, I get the following error. We probably want to handle empty metadata.

image
- Double clicking on the box doesn't collapse it or expand it. Is that happening to you too?

YohannParis:
>     * Double clicking on the box doesn't collapse it or expand it. Is that happening to you too?

@RosaRomeroGomez 
Indeed, this looks fishy, can you tell me if that's what you meant in the code.
I think a merge conflict was not done properly.

YohannParis:
>     * For the SIR model in PNC format, since we don't have metadata, when I click on the SIR box, I get the following error. We probably want to handle empty metadata.

Thank you, this was fixed.

client/src/views/Models/Model.vue
```diff
@@ -366,11 +374,19 @@
     }
 
     onNodeClick (node: GraphNodeInterface): void {
+      // Merge node metadata with Variables metadata. c.f. GraphNodeInterface type
+      const metadata = node.metadata ? node.metadata.flat() : null;
+
+      // Select which tab should be open first, then open the drilldown.
+      this.drilldownActiveTabId = metadata ? 'metadata' : 'parameters';
```
RosaRomeroGomez, line 381: All the metadata that comes from gromets should be displayed in the metadata tab. Then, the user could switch to see the parameters and knowledge tabs that should be populated with data coming from Cosmos. So I don't think this condition is really necessary?

client/src/views/Models/Model.vue
```diff
@@ -351,10 +352,17 @@
       this.selectedLayoutId = layoutId;
     }
 
-    async searchCosmos (keyword: string): Promise<void> {
+    async getDrilldownKnowledge (): Promise<void> {
       try {
-        const response = await cosmosSearch(filterToParamObj({ cosmosQuery: [keyword] }));
-        this.drilldownKnowledge = response;
+        if (this.selectedGraphMetadata?.constructor === Array) {
```
RosaRomeroGomez, line 357: What happens if there is no metadata for a node? I think we used to have an error message informing about this. I think it looks weird and not informative when the panel is just empty.
image

RosaRomeroGomez (approved):
Other than that, this looks good.

client/src/views/Models/Model.vue
YohannParis: There is a message, using the new component you created.
I think you stumble onto a bug from `main`.

I fixed it, so it should display the message as you expected:
Screen Shot 2021-08-02 at 20 04 28

PR #317, opened 2021-07-29
Use aske-id data in models view to populate knowlege drilldown panel with cosmos data

mj3cheun:
note, when testing i noticed that xdd does not seem to have the aske-ids given by the grfn metadata in their system

### DIFF
--- a/client/src/types/typesCosmos.ts
+++ b/client/src/types/typesCosmos.ts
@@ -43,9 +43,10 @@ export interface CosmosSearchObjectsInterface {
 export interface CosmosSearchInterface {
   /** Cosmos API version */
   v: string;
-  total: number;
-  page: number;
-  objects: CosmosSearchObjectsInterface[];
+  total?: number;
+  page?: number;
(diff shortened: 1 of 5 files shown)


client/src/types/typesCosmos.ts
```diff
@@ -43,9 +43,10 @@ export interface CosmosSearchObjectsInterface {
 export interface CosmosSearchInterface {
   /** Cosmos API version */
   v: string;
-  total: number;
-  page: number;
-  objects: CosmosSearchObjectsInterface[];
+  total?: number;
+  page?: number;
+  objects?: CosmosSearchObjectsInterface[];
+  error?: string,
```
YohannParis, line 49: I love it ❤️

YohannParis (approved):
Looks great and smart.

mj3cheun:
Awesome! And it seems that the SIR aske-ids will be added to xDD by late today to tomorrow morning so that resolves the note I made earlier. We can merge this I think.

PR #318, opened 2021-07-29
Odd issues in one PR

### LINKED ISSUE #307: Move add/remove all parameters and variables buttons to the model representation panel
Idea is to consolidate add/remove functionality in one place and to remove redundant buttons.

### LINKED ISSUE #308: Input validation on simulate page
When the user enters parameter values that are invalid (such as “NaN”), there is no message or indicator of why the variables panel doesn’t show anything. 

Prevent "Run" button from working if there are invalid values in the input. Show error messages indicating which input box contains the problematic value.

### LINKED ISSUE #311: Use lower case postfix for numbers
This is mostly an issue for counters. Use lower case postfix (such as k) as per scientific notation.

### LINKED ISSUE #327: Disable horizontal scroll in parameters panel in simulate view
Note: this behaviour seems inconsistent...

Screen Shot 2021-07-30 at 11 53 43 AM

YohannParis:
**What**
- A) fix #311 
- B) fix #307 
- C) fix #308 
- D) fix #327 

**How**
- A) Added a new function in `NumberUtils.ts`, updated the parameters panel and the counter compoment.
- B) Move the Add/Remove all to the Simulation view _Model_ panel, and remove them from the _Parameters_ and _Variables_ panel
- C) Add test and message if one of the parameters values is not a number.
- D) `overflow-x: hidden`
- E) Default `_init` parameters without an available default values is set to 1, so the user can run the model right away without setting up multiple dozens of values by hand.
- clean the `MessageDisplay` component to remove the named slot.

**Screenshot & Videos**
A) 
Screen Shot 2021-07-29 at 16 37 57

B)
Screen Shot 2021-07-30 at 15 13 10

C)

https://user-images.githubusercontent.com/636801/128058133-613a10bd-66ff-4ba9-a16c-692347b6d73b.mov

E) Example with _marm-model_
Screen Shot 2021-08-03 at 11 19 54


RosaRomeroGomez:
I noticed this error when we get not valid default values for parameters. We can log it and address it later.
image

RosaRomeroGomez (approved):

Issue #314, opened 2021-07-29
Make line charts half the size so we can fit more of them without having to scroll


RosaRomeroGomez:
Maybe we can reduce the height and the labels size a bit?

mj3cheun:
There will be a limit to how much the height of the charts can be reduced without being illegible. In addition, the number of labels (on the axis) will probably need to be reduced in order to prevent crowding at smaller graph sizes.

Issue #303, opened 2021-07-27
Fix crash when multiple runs are specified and run (for the first time) all at once


mj3cheun:
in simulate view

Issue #295, opened 2021-07-26
Allow parameters panel to scroll down in the simulate view


mj3cheun:
Currently the parameters panel overflow is hidden when there are more parameters then can fit on the screen. Allow the panel to scroll to allow access to all parameters regardless of the number shown. In addition, fix the overflow when large titles are encountered.

Test using marm model.

Issue #292, opened 2021-07-26
Provide more feedback to the user on saved run in simulate


mj3cheun:
Currently there is little feedback provided when a new run is saved using the "bookmark button" in the simulate page and there is none when the button is pressed using touch click.

One way to provide feedback include changing the icon to a checkmark on click

<img width="290" alt="Screen Shot 2021-07-26 at 12 46 03 PM" src="https://user-images.githubusercontent.com/14062458/127027208-960ee541-d3ac-41af-ab06-85de05f3163b.png">

RosaRomeroGomez:
I think we actually would rather the other option using an outline. However, instead of using a blue outline we could use the yellow selection color.
![image](https://user-images.githubusercontent.com/10552785/127028893-5dcfcc9a-d147-42a5-8014-1c7a56b11f62.png)

PR #320, opened 2021-07-29
Consolidated fixes 1

mj3cheun:
Solves the following issues:

- #289 
- #292
- #295
- #303
- #304
- #314

### DIFF
--- a/client/src/components/widgets/charts/MultiLinePlot.vue
+++ b/client/src/components/widgets/charts/MultiLinePlot.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="mb-2" ref="container">
+  <div ref="container">
     <div class="header" :style="{color: strokeColor}">
       {{title}}
       <slot/>
@@ -54,7 +54,7 @@
     }
 
     mounted (): void {
-      this.width = 500;
+      this.width = 1000;
       this.height = 150;
       this.margin = DEFAULT_CONFIG.margin;
       this.refresh();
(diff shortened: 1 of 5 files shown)


YohannParis (approved):
Looks fine and dandy

RosaRomeroGomez:
Some comments:

- I think the goal for the save button was to make it look pressed or in other words, I think this button should actually act as a checkbox or toggle. So if it is checked (and it looks pressed), your runs are saved and displayed but if it is not, they are not saved and displayed. Let me know if this makes sense. 
- For line charts, I think we need to display fewer labels. d3 should allow you to define the interval for them.

YohannParis:
>     * I think the goal for the save button was to make it look pressed or in other words, I think this button should actually act as a checkbox or toggle. So if it is checked (and it looks pressed), your runs are saved and displayed but if it is not, they are not saved and displayed. Let me know if this makes sense.

This is a completely new behaviour. I don't think it will mix well with the _Auto-run_ option, suddenly at every change of a parameters we save an extra run? This can create dozens of runs really fast.

mj3cheun:
Let me know which save button behaviour you would prefer, I think either could work, although for the toggle behaviour I would be tempted to incorporate an actual checkbox into the button.

As for displaying fewer labels, I think that could be done in a separate PR as part of a larger task to make the line charts truly responsive.

RosaRomeroGomez (approved):

PR #329, opened 2021-07-30
[FEATURE] Group models by name

adamocarolli:
### Description
Group Donu models that have the same name together. For example if a FunctionalNetwork type model and a PetriNetClassic type model had the same name they'd be grouped together as the same model with different representations.

### Changes
- Remove hard-coding of model groupings. Group only based on model name being identical.
- Remove hard-coding of model preview image positions, instead select screenshot based on matching model names.

### Considerations
- As discussed in comment grouping model's by their names is inherently flawed. There is no guarantee that a model with the same name will be the same model.
- Some models like SIR are the same model but have different names, this strategy currently fails for those use cases. I've chosen to leave these mismatches in for now as models are being added to Donu and then before the demo we can include a filter list to correct.

### Testing
- No special instructions.


RosaRomeroGomez (approved):
Thanks! We'll need to update the screenshots when we get a finalized version of everything but this solves two issues actually :)

PR #331, opened 2021-08-02
Removed pointer from control for container nodes

RosaRomeroGomez:
**What**

- Removed pointer from the control added to collapsible nodes (container nodes). The container itself allows expanding and collapsing already so the pointer in the control is unneeded.
- Added the pointer to the container nodes so it is clearer that they can be collapsed and expanded.

**Testing**

- Select a model
- Try to expand and collapse container nodes. A cursor should show up for them but not for the control.

https://user-images.githubusercontent.com/10552785/127851722-ae34f01d-1f69-490a-84bc-eb16a838753b.mov

### DIFF
--- a/client/src/graphs/svg/renderers/EpiRenderer.ts
+++ b/client/src/graphs/svg/renderers/EpiRenderer.ts
@@ -130,7 +130,7 @@ export default class EpiRenderer extends SVGRenderer {
           .style('stroke', DEFAULT_STYLE.node.stroke)
           .style('stroke-width', EpiRenderer.calcNodeStrokeWidth)
           .style('stroke-dasharray', EpiRenderer.calcNodeStrokeStyle)
-          .style('cursor', d => (d as any).nodes ? '' : 'pointer');
+          .style('cursor', 'pointer');
       } else {
         selection.append('ellipse')
           .attr('cx', d => (d as any).width * 0.5)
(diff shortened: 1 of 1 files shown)


mj3cheun (approved):

mj3cheun:
Actually what I had in mind more for a fix is to make only the top-right indicator box clickable and to make the rest of the collapsible node unclickable, so basically the exact opposite of whats happening here. Here are a few reasons why I think this is a good idea:

- The main reason why is because by making the collapsible node clickable, it increases the chance a node will be accidentally collapsed and disrupt the user's flow. By restricting the click to the top-right indicator, the chances this will happen should drastically reduce.
- People using windows/mac os are already used to clicking on that top-right to collapse/expand windows, they already have that mental model

disadvantages of doing it this way is:

- It makes collapsing nodes, particularly multiple nodes in quick succession, considerably slower and possibly more awkward

I will leave final decision to you, the changes you made do work so if you want to stick to it, looks good

RosaRomeroGomez:
> Actually what I had in mind more for a fix is to make only the top-right indicator box clickable and to make the rest of the collapsible node unclickable, so basically the exact opposite of whats happening here. Here are a few reasons why I think this is a good idea:
> 
> * The main reason why is because by making the collapsible node clickable, it increases the chance a node will be accidentally collapsed and disrupt the user's flow. By restricting the click to the top-right indicator, the chances this will happen should drastically reduce.
> * People using windows/mac os are already used to clicking on that top-right to collapse/expand windows, they already have that mental model
> 
> disadvantages of doing it this way is:
> 
> * It makes collapsing nodes, particularly multiple nodes in quick succession, considerably slower and possibly more awkward
> 
> I will leave final decision to you, the changes you made do work so if you want to stick to it, looks good

Thank you @mj3cheun for the explanation. I totally agree with you that the best option would be to disable the clicking for the parent container and only allowing it for the control. There is a reason the control is there and the first intuition for users when they see it is probably clicking on it. However, I'm not sure how much control we have over disabling it in the parent. This is not a priority but I'll give it a look when I have some spare time. In the meantime, I'll keep this open.

RosaRomeroGomez:
Going to close this for now since we don't have time to address this.

PR #332, opened 2021-08-02
Coordination between panels in the Simulation space

RosaRomeroGomez:
**What**
- Added coordinated highlighting from Models panel to Vars/Params panel.

**How**
- Added border for parameters and variables containers if they have the same name than the selected node in the Model panel. Ideally, this should use the `uid` but it doesn't seem to match. NOTE: We probably want to double check that in the parser and adapt it. 

**Testing**

- Select a model and open the simulation view
- Add parameters and run the model.
- Select nodes from the Model visualization. The corresponding containers in the params/variable panels should get highlighted.

https://user-images.githubusercontent.com/10552785/127869001-fd07e022-c4de-45ff-b03b-2fd17c040560.mov

### DIFF
--- a/client/src/views/Simulation/components/ParametersPanel.vue
+++ b/client/src/views/Simulation/components/ParametersPanel.vue
@@ -59,7 +59,7 @@
           class="parameter"
           v-for="(parameter, index) of displayedParameters"
           :key="index"
-          :class="[{ hidden: parameter.hidden, highlighted: parameter.metadata.name === highlighted }]"
+          :class="{ hidden: parameter.hidden, highlighted: parameter.metadata.name === highlighted }"
         >
           <h4 :title="parameter.metadata.name">{{ parameter.metadata.name }}</h4>
           <input type="text" v-model.number="parameterValues[parameter.uid]" />
--- a/client/src/views/Simulation/components/ParametersPanel.vue
+++ b/client/src/views/Simulation/components/ParametersPanel.vue
@@ -59,7 +59,7 @@
           class="parameter"
           v-for="(parameter, index) of displayedParameters"
           :key="index"
-          :class="{ hidden: parameter.hidden }"
+          :class="[{ hidden: parameter.hidden, highlighted: parameter.metadata.name === highlight }]"
         >
           <h4 :title="parameter.metadata.name">{{ parameter.metadata.name }}</h4>
           <input type="text" v-model.number="parameterValues[parameter.uid]" />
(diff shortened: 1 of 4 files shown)


client/src/views/Simulation/components/ParametersPanel.vue
```diff
@@ -59,7 +59,7 @@
           class="parameter"
           v-for="(parameter, index) of displayedParameters"
           :key="index"
-          :class="{ hidden: parameter.hidden }"
+          :class="[{ hidden: parameter.hidden, highlighted: parameter.metadata.name === highlighted }]"
```
YohannParis, line 62: You do not need to use an array, the object is enough:
```
:class="{
  hidden: parameter.hidden,
  highlighted: parameter.metadata.name === highlighted,
}"
```

client/src/views/Simulation/components/ParametersPanel.vue
```diff
@@ -103,6 +103,7 @@
   export default class ParametersPanel extends Vue {
     @Prop({ default: false }) expanded: boolean;
     @Prop({ default: null }) modelId: number;
+    @Prop({ default: '' }) highlighted: string;
```
YohannParis, line 106: Why not `null` in this case? Just curious.

client/src/views/Simulation/components/ParametersPanel.vue
```diff
@@ -403,6 +404,10 @@
   .parameter.hidden {
     opacity: 0.5;
   }
+
+  .parameter.highlighted {
+    border: 1px var(--selection) solid;
```
YohannParis, line 409: It will be better to add a border to parameter to avoid the jump when being highlighted:

https://user-images.githubusercontent.com/636801/127937403-34d311a5-ca2d-4a18-9cea-69132de1354b.mov

doing something like this:
```CSS
.parameter {
  ...
  border: 1px solid transparent;
}

.parameter.highlighted {
  border-color: var(--selection);
}
```

YohannParis (approved):
Minor comments on `ParametersPanel.vue` that also applies to `VariablesPanel.vue`.

client/src/views/Simulation/components/ParametersPanel.vue
RosaRomeroGomez: Thank you. Done.

client/src/views/Simulation/components/ParametersPanel.vue
RosaRomeroGomez: Not sure what's the convention we follow for default values actually. I'll change it to `null`

client/src/views/Simulation/components/ParametersPanel.vue
RosaRomeroGomez: Great point, changed.

YohannParis (approved):

mj3cheun (approved):

Issue #328, opened 2021-07-30
Fix yellow dots not coordinated with the Parameters and Variables Panel 


emilykuang:
Steps to reproduce:
1. Add all variables and parameters using + on the parameters panel
2. Remove variables and parameters individually by double-clicking on model nodes (seems to be a delay in hiding the yellow dots from nodes) 

Screenshot shows S and R still having a yellow circle with nothing in the parameter and variable panels: 
![image](https://user-images.githubusercontent.com/23723378/127682399-dea4ef83-1cff-4558-8440-becfb24bb282.png)

PR #333, opened 2021-08-03
Remove hidden property

### LINKED ISSUE #281: [Simulation] Remove `hidden` property in favour of `edited`.
In the simulation view _parameters_ and _variables_ have two different properties `hidden` and `edited` who behaves differently.

1. **`hidden` needs to be removed, as it's out of scope for now.**

2. **`edited` needs to be kept, and be renamed `hidden`.**
   This part is more delicate has `edited` is a positive test, and `hidden` is a negative one.

emilykuang:
**What**

- fix #281 by removing the `hidden` property of parameters and variables and renaming `edited` property to `displayed`
- `displayed` makes more sense than `hidden` since it was the opposite meaning

**Testing**

- Select a model and open the simulation view 
- Add parameters using the plus sign and make sure the yellow dots appear in the Model Panel
- Remove parameters using the no icon and make sure the yellow dots disappear 

Screenshot:
image

Updated screenshot with highlight box:
image

### DIFF
--- a/client/src/views/Simulation/components/ModelPanel.vue
+++ b/client/src/views/Simulation/components/ModelPanel.vue
@@ -15,7 +15,7 @@
     <global-graph
       v-if="graph"
       :data="graph"
-      :edited-nodes="editedNodes"
+      :edited-nodes="displayedNodes"
       @node-dblclick="onNodeDblClick"
     />
   </section>
@@ -64,15 +64,15 @@
       return this.getSimModel(this.model.id).variables;
     }
 
-    get editedNodes (): Graph.SubgraphNodeInterface[] {
+    get displayedNodes (): Graph.SubgraphNodeInterface[] {
       const nodes = this.graph?.nodes;
       if (nodes) {
         const highlightedLabels = [
           ...this.parameters
-              .filter(parameter => parameter.edited)
+              .filter(parameter => parameter.displayed)
               .map(parameter => parameter.metadata.name),
           ...this.variables
-              .filter(variable => variable.edited)
+              .filter(variable => variable.displayed)
               .map(variable => variable.metadata.name),
         ];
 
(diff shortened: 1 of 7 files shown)


client/src/views/Simulation/components/ModelPanel.vue
```diff
@@ -15,7 +15,7 @@
     <global-graph
       v-if="graph"
       :data="graph"
-      :edited-nodes="editedNodes"
+      :edited-nodes="displayedNodes"
```
YohannParis, line 18: you might change the name of the prop as well from `edited-nodes` to `displayed-nodes` on the `<global-graph>` component.

YohannParis (approved):
Looks good!

client/src/views/Simulation/components/ModelPanel.vue
emilykuang: Thanks for the suggestion, updated.

YohannParis (approved):

RosaRomeroGomez:
The yellow dots are not disappearing for me when I unselect one of the parameters or variables from the Model panel. I suspect it is related to the bug you logged here: https://github.com/uncharted-aske/HMI/issues/328 Are you guys experiencing the same issue? @YohannParis @mj3cheun ?

RosaRomeroGomez (approved):

RosaRomeroGomez:
> > The yellow dots are not disappearing for me when I unselect one of the parameters or variables from the Model panel. I suspect it is related to the bug you logged here: #328 Are you guys experiencing the same issue? @YohannParis @mj3cheun ?
> 
> Yes, I noticed this issue too before I started refactoring. The dots appear and disappear as expected when you only use the buttons at the top or only by double-clicking on the nodes. However, if you try to add all dots using the + button, then try to remove one by double-clicking, the dot doesn't disappear.

Thanks @emilykuang for the clarifications. I think this PR seems fine. Maybe you can assign yourself that bug and see if you can find the issue?

PR #334, opened 2021-08-03
Change from dropdown to modal box

### LINKED ISSUE #296: Change run config dropdown to a modal in the simulate view
Take all run config functionality (start time, stop time, step size, autorun) and put it in a model
Update run config button to reflect the change

YohannParis:
**What**
- fix #296 

**How**
- Remove the _dropdown_ element and use the _Modal_ component.

**Video**

https://user-images.githubusercontent.com/636801/128081253-9ac729e6-f31f-42f2-9296-f3162fba8a57.mov

### DIFF
--- a/client/src/views/Simulation/components/RunButton.vue
+++ b/client/src/views/Simulation/components/RunButton.vue
@@ -1,38 +1,35 @@
 <template>
-  <div class="btn-group">
+  <div class="run-button btn-group">
     <button
       type="button"
       class="btn btn-primary blue"
       :disabled="disabled"
       @click="onClickRun"
     >
-      <font-awesome-icon :icon="['fas', ( autoRunInput ? 'pause' : 'play') ]" />
+      <font-awesome-icon :icon="['fas', displayIcon ]" />
       {{ displayText }}
     </button>
     <button
       type="button"
-      class="btn btn-primary blue dropdown-toggle dropdown-toggle-split"
-      :class="{ 'open': dropdownOpen }"
-      @click="dropdownOpen = !dropdownOpen"
-    />
-    <div class="dropdown-menu dropdown-menu-right" :class="{ 'show': dropdownOpen }">
-      <input type="checkbox" v-model="autoRunInput" id="auto-run-input" />
-      <label
-        for="auto-run-input"
-        title="Enable model to run automatically on when Parameters are updated">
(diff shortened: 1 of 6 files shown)


RosaRomeroGomez:
A couple of comments/suggestions:

- Maybe we can add alt text or tooltip for the `Run` and `Configuration` buttons since we don't have labels there and that will be helpful for the user (we already have them for the save and reset buttons). 
- I was thinking that the modal is quite small. Maybe we can put everything on the same row or have two columns, one for the auto-run checkbox and one for the rest. I was doing some testing in terms of width (don't pay attention to the layout) and I think this might be better. I'm using `width: 40vw`.
image

YohannParis:
>Maybe we can add alt text or tooltip for the Run and Configuration buttons since we don't have labels there and that will be helpful for the user (we already have them for the save and reset buttons).

I added a `tooltip` css feature to display tooltip that are not the `title` attribute.

https://user-images.githubusercontent.com/636801/128211633-5e90efac-e281-4cc3-a9b3-c21464e93f26.mov

YohannParis:
> I was thinking that the modal is quite small. Maybe we can put everything on the same row or have two columns, one for the auto-run checkbox and one for the rest. I was doing some testing in terms of width (don't pay attention to the layout) and I think this might be better. I'm using width: 40vw.

I updated the modal to be wider, to be around 40vw.

Screen Shot 2021-08-04 at 12 55 53

client/src/views/Simulation/components/RunButton.vue
```diff
@@ -1,38 +1,64 @@
...
     </div>
+
+    <!-- Modal to configure Run configuration -->
+    <modal v-if="settingOpen" @close="settingOpen = false">
+      <h5 class="header" slot="header">Model Execution Configuration</h5>
+      <div slot="body" class="run-config">
+        <section>
+          <label for="auto-run-input">Auto-Run</label>
+          <input type="checkbox" v-model="autoRunInput" id="auto-run-input" />
+          <p class="info">Enable model to run automatically when Parameters are updated.</p>
```
RosaRomeroGomez, line 50: should be just `parameters`, no caps.

RosaRomeroGomez (approved):
Looks great, thanks!

PR #336, opened 2021-08-04
Variable band

### LINKED ISSUE #324: Implement band for previous runs instead of dots for line chart

mj3cheun:
Adds band to variable panel and fixes incorrect aggregate function (mean avg value calculated incorrect).

Note: when testing keep in mind that the bands are only enabled when there are more than 5 runs

Fixes #324

Screen Shot 2021-08-04 at 5 17 18 PM

### DIFF
--- a/client/src/components/widgets/charts/MultiLinePlot.vue
+++ b/client/src/components/widgets/charts/MultiLinePlot.vue
@@ -38,6 +38,7 @@
   export default class MultiLinePlot extends Vue {
     @Prop({ default: 'Title' }) title: string;
     @Prop({ default: () => [[]] }) data: any;
+    @Prop({ default: () => [] }) polygon: any;
     @Prop({ default: () => [] }) styles: any;
 
     width: number;
@@ -160,6 +161,17 @@
           .attr('r', 3);
     }
 
+    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+    drawPolygon (svg, data): void {
+      svg.selectAll('polygon')
+        .data([data])
+        .enter().append('polygon')
+          .attr('points', d =>
+              d.map(d => [this.xScale(d.x), this.yScale(d.y)].join(',')).join(' '))
+          .attr('fill', '#647EA9')
+          .attr('opacity', '0.4');
+    }
+
     refresh (): void {
       const currentDataHash = JSON.stringify(this.data) + JSON.stringify(this.styles);
       if (currentDataHash === this.dataHash) {
@@ -178,6 +190,10 @@
 
       svg.append('g')
         .call(this.yAxis);
+
+      if (this.polygon?.length) {
+        this.drawPolygon(svg, this.polygon);
+      }
 
       this.data.map((dataset, i) => this.drawDataset(svg, dataset, this.styles[i]));
 
--- a/client/src/components/widgets/charts/MultiLinePlot.vue
+++ b/client/src/components/widgets/charts/MultiLinePlot.vue
@@ -23,6 +23,10 @@
     label: {
       text: Colors.LABELS.LIGHT,
     },
+    polygon: {
+      fill: '#647EA9',
+      opacity: '0.4',
+    },
   };
 
   const DEFAULT_CONFIG = {
@@ -168,8 +172,8 @@
         .enter().append('polygon')
           .attr('points', d =>
               d.map(d => [this.xScale(d.x), this.yScale(d.y)].join(',')).join(' '))
-          .attr('fill', '#647EA9')
-          .attr('opacity', '0.4');
+          .attr('fill', DEFAULT_STYLE.polygon.fill)
+          .attr('opacity', DEFAULT_STYLE.polygon.opacity);
     }
 
     refresh (): void {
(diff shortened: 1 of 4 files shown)


client/src/components/widgets/charts/MultiLinePlot.vue
```diff
@@ -160,6 +161,17 @@
...
     }
 
+    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+    drawPolygon (svg, data): void {
+      svg.selectAll('polygon')
+        .data([data])
+        .enter().append('polygon')
+          .attr('points', d =>
+              d.map(d => [this.xScale(d.x), this.yScale(d.y)].join(',')).join(' '))
+          .attr('fill', '#647EA9')
```
RosaRomeroGomez, line 171: Could we move this color to the `DEFAULT_STYLE` object?

client/src/views/Simulation/components/VariablesPanel.vue
```diff
@@ -114,10 +115,10 @@
 
   const AGGREGATE_STYLE = {
     node: {
-      fill: Colors.NODES.AGGREGATE,
+      fill: '#EBCB8B',
```
RosaRomeroGomez, line 118: Why this change?

client/src/views/Simulation/components/VariablesPanel.vue
```diff
@@ -164,20 +159,40 @@
...
       variables.map(variable => {
         variable.styles = variable.styles || [];
-        for (let i = 0; i < variable.values.length; i++) {
-          if (variable.values.length > 5) {
-            i !== variable.values.length - 1
-              ? variable.styles.push(mergeStyle(OTHER_STYLE, NO_LINE_STYLE))
-              : variable.styles.push(mergeStyle());
-          } else {
-            variable.styles.push(mergeStyle(i !== variable.values.length - 1 && OTHER_STYLE));
+        if (variable.values.length > 5) {
```
RosaRomeroGomez, line 188: I think we should have this arbitrary number of runs as a constant in this panel. Something like `THRESHOLD = 5`

client/src/store/modules/simulation.ts
```diff
@@ -246,7 +246,7 @@ const mutations: MutationTree<HMI.SimulationState> = {
       // A variable.values are a list of runs each containing
       // a list of {x: step, y: value} for each step.
       const valuesOfAllRunsPerStep = {};
-      for (let run = 0; run < state.numberOfSavedRuns; run++) {
+      for (let run = 0; run < variable.values.length; run++) {
```
YohannParis, line 249: That would explain the wrong value. Cheers!

client/src/views/Simulation/components/VariablesPanel.vue
```diff
@@ -164,20 +159,40 @@
     @Action hideAllVariables;
     @Action showAllVariables;
 
+    calcPolygon (variable: HMI.SimulationVariable): void {
```
YohannParis, line 162: I understand that for d3 this data is a Polygon, but for us this is a band of values.
Maybe I'm picky but it would be clearer if the function and variables were named after what it represent.
Unless you thought of another situation were we can use polygons?

client/src/views/Simulation/components/VariablesPanel.vue
```diff
@@ -164,20 +159,40 @@
...
+      const numRuns = variable.values.length;
```
YohannParis, line 163: You only use `numRuns - 1`, so why not just set it as `const numRuns = variable.values.length - 1;`?

client/src/views/Simulation/components/VariablesPanel.vue
```diff
@@ -164,20 +159,40 @@
...
+        run.forEach((point, pointIndex) => {
+          xArr[pointIndex] = point.x;
+          yMinArr[pointIndex] = Math.min(yMinArr[pointIndex] ?? point.y, point.y);
+          yMaxArr[pointIndex] = Math.max(yMaxArr[pointIndex] ?? point.y, point.y);
+        });
+      }
+
+      variable.polygon = [
+        ...yMinArr.map((y, i) => ({ x: xArr[i], y })),
+        ...yMaxArr.map((y, i) => ({ x: xArr[i], y })).reverse(),
```
YohannParis, line 179: Could you add a small comment explaining why you use `reverse()`, it's not obvious right away.

YohannParis (commented):
Looks great, and it's really neat to see the band.

client/src/views/Simulation/components/VariablesPanel.vue
mj3cheun: `Colors.NODES.AGGREGATE` has been deleted at some point so the colour code is no longer there causing a problem where the aggregate run looks the same as the current run

client/src/views/Simulation/components/VariablesPanel.vue
mj3cheun: I would prefer to keep the word `polygon` there in some way because it is important in understanding why the data structure is structured the way it is. I see your point though. What do you think of using the fn name `calcBandPolygon`?

YohannParis:
Thank you for al your changes. Especially that comment! It is above and beyond.

RosaRomeroGomez (approved):
Looks good, thanks!

PR #337, opened 2021-08-04
Ability to show metadata for PNC EMMAA GroMEts

RosaRomeroGomez:
**What**

- Added two additional types of metadata included in PNC EMMAA GroMETs (`IndraAgentReferenceSet` and `ReactionReference`)
- Using this metadata, we enable the ability to get agent references and pieces of evidence from EMMAA (similarly to how we do it for the Graphs).

**Testing**
- Select one of the two EMMAA gromets (MARM model or RAS model).I'd highly recommend you to test this out using the MARM model since the current RAS model is too big.
- Select a node from the model (edges don't include metadata for gromets)

image
image

NOTES: Testing might be annoying because you need to use the current EMMAA GroMEt visualizations, which take some time to load and are basically a hard ball, which makes difficult to select things. I also noticed, we don't seem to be getting DOIs from the EMMAA evidence so I wasn't able to leverage COSMOS for that. I need to follow up with HMS. I also noticed a bunch of inconsistent CORS errors by using these EMMAA endpoints. Not sure what's going on there, maybe @mj3cheun would be able to help out with this.

### DIFF
--- a/client/src/views/Models/Model.vue
+++ b/client/src/views/Models/Model.vue
@@ -382,11 +382,10 @@
       this.drilldownMetadata = node.metadata ? node.metadata.flat() : null;
 
       // Select tab to open
-      this.drilldownActiveTabId = this.drilldownMetadata.metadata_type === GroMEt.MetadataType.ReactionReference ? 'knowledge' : 'metadata';
-
+      // this.drilldownActiveTabId = this.drilldownMetadata[0].metadata_type === GroMEt.MetadataType.ReactionReference ? 'knowledge' : 'metadata';
       this.drilldownPaneSubtitle = `${node.nodeType} (${node.dataType})`;
       this.drilldownPaneTitle = node.label;
-      this.getDrilldownKnowledge();
+      // this.getDrilldownKnowledge();
     }
 
     onBackgroundClick ():void {
--- a/client/src/views/Models/Model.vue
+++ b/client/src/views/Models/Model.vue
@@ -79,6 +79,7 @@
         v-if="drilldownActiveTabId === 'metadata' && drilldownMetadata"
         slot="content"
         :metadata="drilldownMetadata"
+        :model-name="selectedModel.name"
         @open-modal="onOpenModalMetadata"
       />
       <parameters-pane
@@ -381,11 +382,9 @@
       // Merge node metadata with Variables metadata. c.f. GraphNodeInterface type
       this.drilldownMetadata = node.metadata ? node.metadata.flat() : null;
 
-      // Select tab to open
-      // this.drilldownActiveTabId = this.drilldownMetadata[0].metadata_type === GroMEt.MetadataType.ReactionReference ? 'knowledge' : 'metadata';
       this.drilldownPaneSubtitle = `${node.nodeType} (${node.dataType})`;
       this.drilldownPaneTitle = node.label;
-      // this.getDrilldownKnowledge();
+      this.getDrilldownKnowledge();
     }
 
     onBackgroundClick ():void {
--- a/client/src/views/Models/Model.vue
+++ b/client/src/views/Models/Model.vue
@@ -365,6 +365,7 @@
             }));
           }
         }
+
       } catch (e) {
         throw Error(e);
       }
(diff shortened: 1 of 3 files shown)


client/src/views/Models/Model.vue
```diff
@@ -376,8 +377,6 @@
     }
 
     onNodeClick (node: GraphNodeInterface): void {
-      // Select which tab should be open first, then open the drilldown.
-      this.drilldownActiveTabId = 'metadata';
```
YohannParis, line 380: Why removing it? Don't we want the panel to always start with the _Metadata_ panel?

client/src/views/Models/components/DrilldownPanel/MetadataPane.vue
```diff
@@ -78,27 +115,65 @@
...
   import { formatFullDateTime } from '@/utils/DateTimeUtil';
   import MessageDisplay from '@/components/widgets/MessageDisplay.vue';
+  import LoadingAlert from '@/components/widgets/LoadingAlert.vue';
 
   const METADATA_TYPES_ORDER = [
     GroMET.MetadataType.TextDefinition,
     GroMET.MetadataType.TextParameter,
     GroMET.MetadataType.EquationDefinition,
     GroMET.MetadataType.CodeSpanReference,
+    GroMET.MetadataType.CodeSpanReference,
```
YohannParis, line 137: Looks like a double `GroMET.MetadataType.CodeSpanReference` here.

client/src/views/Models/components/DrilldownPanel/MetadataPane.vue
```diff
@@ -78,27 +115,65 @@
...
   export default class MetadataPane extends Vue {
     @Prop({ default: [] }) metadata: GroMET.Metadata[];
+    @Prop({ default: '' }) modelName: string;
+
+    isLoading = false;
+    agentsReferences: EMMAA.EmmaaEntityInfoInterface[] = [];
+    nodeEvidences: EMMAA.EmmaaEvidenceEvidenceInterface[] = [];
+
+    @Watch('metadata') onDataChange (): void {
+      if (this.metadata.find(datum => datum.metadata_type === GroMET.MetadataType.IndraAgentReferenceSet)) {
```
YohannParis, line 157: You can simplify your tests inside `onDataChange ()` and `mounted ()` by using the new methods you wrote `isTypeIndraAgentReferenceSet` and `isTypeReactionReference`.
That would keep the `metadata_type` test in one place in case we have to update them.

```JS
// Fom:
this.metadata.find(datum => datum.metadata_type === GroMET.MetadataType.IndraAgentReferenceSet)

// To:
this.metadata.find(this.isTypeIndraAgentReferenceSet)
```

client/src/views/Models/components/DrilldownPanel/MetadataPane.vue
```diff
@@ -65,6 +66,42 @@
...
+          </details>
+        </div>
+      </template>
+
+      <template v-if="isTypeReactionReference(datum)">
+        <summary :title="datum.uid">Reaction Reference(s)</summary>
+            <details v-if="hasNodeEvidences" class="metadata" open>
+              <summary>Evidence ({{nodeEvidences.length}})</summary>
+              <div class="metadata-content">
+                <div class="metadata-content">
```
YohannParis, line 93: Why the double `<div class="metadata-content">` ?

YohannParis (commented):
Looks really well organize.

client/src/views/Models/components/DrilldownPanel/MetadataPane.vue
```diff
@@ -78,27 +115,65 @@
...
+      if (this.metadata.find(datum => datum.metadata_type === GroMET.MetadataType.IndraAgentReferenceSet)) {
+        this.fetchAgentsReferences();
+      }
+      if (this.metadata.find(datum => datum.metadata_type === GroMET.MetadataType.ReactionReference)) {
+        this.fetchNodeEvidences();
+      }
+    }
+
+    get hasNodeEvidences (): boolean {
+      return this.nodeEvidences.length > 0;
```
YohannParis, line 175: I got the following error while testing:
Screen Shot 2021-08-04 at 20 04 04

Maybe on line 251 you can make sure `this.nodeEvidences` is an array by adding conditional test `?? [];`

client/src/views/Models/Model.vue
RosaRomeroGomez: I thought we already initialized it above but you're right, I'll put it back.

client/src/views/Models/components/DrilldownPanel/MetadataPane.vue
RosaRomeroGomez: Good catch.

client/src/views/Models/components/DrilldownPanel/MetadataPane.vue
RosaRomeroGomez: Great point, thanks!

client/src/views/Models/components/DrilldownPanel/MetadataPane.vue
RosaRomeroGomez: I think it was just a mistake. Good catch.

client/src/views/Models/components/DrilldownPanel/MetadataPane.vue
RosaRomeroGomez: Done.

client/src/views/Models/Model.vue
YohannParis: It is initialized at the beginning. But if you remove this it will not open on the metadata panel each time you click on a node.

PR #339, opened 2021-08-05
Models legend

### LINKED ISSUE #291: Add legend to simulate page to clarify the colour encodings used by the panels
white is current run
blue is previous runs
orange is aggregate

### LINKED ISSUE #298: Add legend to the Models view to clarify visual encodings for GroMets
Legend is collapsed by default

image

Displayed legend
image

NOTE: Latest version
image

YohannParis:
**What**
 - fix #298

**How**
- [x] Create a collapsible _GraphLegend_ component.
- [x] Add the _GraphLegend_ to the _Model_ panel on the Simulation page.
- [x] Fill the content using the same SVG as the _EpiRenderer_, and start using CSS custom properties to style the legend.

**Video**

https://user-images.githubusercontent.com/636801/128879241-174f3f29-76b3-44c2-a53e-f6eed35d7f88.mov

### DIFF
--- a/client/src/views/Models/components/Graphs/GraphLegend.vue
+++ b/client/src/views/Models/components/Graphs/GraphLegend.vue
@@ -89,9 +89,10 @@
     --background-colour: white;
     background-color: var(--background-colour);
     border: var(--border);
-    border-radius: 5px 5px 0 0;
-    bottom: 0;
-    left: 1rem;
+    border-radius: 5px;
+    bottom: .5em;
+    left: 1em;
+    overflow: hidden;
     position: absolute;
   }
 
--- a/client/src/views/Models/components/Graphs/GraphLegend.vue
+++ b/client/src/views/Models/components/Graphs/GraphLegend.vue
@@ -138,7 +138,7 @@
   .graph-legend .legend svg {
     display: inline-block;
     margin-right: 1em;
-    width: 8em;
+    width: 5em;
   }
 
   .graph-legend .legend rect,
--- a/client/src/views/Models/components/Graphs/GraphLegend.vue
+++ b/client/src/views/Models/components/Graphs/GraphLegend.vue
@@ -1,30 +1,84 @@
 <template>
-  <aside class="graph-legend">
-    <header>Legend</header>
-    <ul>
-      <li>Container</li>
-      <li>Parameter</li>
-      <li>Initial Condition</li>
-      <li>Variable</li>
-      <li>Other</li>
-    </ul>
-    <span>Sensitivity score (0&ndash;1)</span>
+  <aside class="graph-legend" :class="{ 'open' : isOpen }">
+    <header>
+      Legend
+      <button type="button" @click="toggleLegend()">
+        <font-awesome-icon :icon="['fas', toggleIcon ]" />
+      </button>
+    </header>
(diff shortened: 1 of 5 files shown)


client/src/views/Models/components/Graphs/GraphLegend.vue
```diff
@@ -0,0 +1,168 @@
...
+    padding: 1em;
+    transition: max-height 250ms, padding 50ms;
+  }
+</style>
+<style>
+  /* For SVG you cannot scope the <style> */
+  .graph-legend .legend svg {
+    display: inline-block;
+    margin-right: 1em;
+    width: 8em;
```
emilykuang, line 141: I prefer the legend to be a bit smaller and played with values from 4em to 6em. Smaller values will make the text inside the nodes hard to read. I'll let @RosaRomeroGomez make the call.

emilykuang (approved):
Looks great! I like that the legend is open by default when I select the model and then closed by default in the simulation view since the model panel is smaller.

RosaRomeroGomez:
Looks really good now we have a smaller legend. Great job @YohannParis . Not really relevant but I was expecting you to use d3 to generate the legend instead of actually creating the svg elements and hard-code them. But I think it is good for now. I just feel the component would be easier to reuse if we used d3.

client/src/views/Models/components/Graphs/GraphLegend.vue
```diff
@@ -0,0 +1,168 @@
...
+          Container
+        </li>
+        <li class="legend-parameter">
+          <svg viewBox="-5 -5 130 50">
+            <rect x="40" y="5" :rx="DEFAULT_STYLE.node.borderRadius" width="30" height="30"></rect>
+          </svg>
+          Parameter
+        </li>
+        <li class="legend-initial-condition">
+          <svg viewBox="-5 -5 130 50">
```
RosaRomeroGomez, line 34: Maybe we can have this viewBox dimensions as a constant so we don't need to repeat them every time?

client/src/views/Models/components/Graphs/GraphLegend.vue
```diff
@@ -0,0 +1,168 @@
...
+            <ellipse cx="60" cy="20" rx="60" ry="20"></ellipse>
+            <text x="60" y="25">Node IC</text>
```
RosaRomeroGomez, line 36: The same for the text coordinates.

RosaRomeroGomez (approved):

YohannParis:
> Looks really good now we have a smaller legend. Great job @YohannParis . Not really relevant but I was expecting you to use d3 to generate the legend instead of actually creating the svg elements and hard-code them. But I think it is good for now. I just feel the component would be easier to reuse if we used d3.

That comes back to our discussion; I wanted to leverage `EpiRenderer` to not use hard-coded values. But that will involve a lot of refactoring to make the `EpiRenderer` more versatile.
So I used the same values that the renderer used and made the SVGs.

I did not use d3 because there is no data in this legend. I mean, I could create fake data and make the whole legend a big SVG, but that will be overcomplicated without the refactoring mentioned above.

I did make the changes you requested by removing all hard-coded values and use only the one used on the _Model_ view.

PR #341, opened 2021-08-05
[FEATURE] Add text search query to model's list view

adamocarolli:
### Description
Add Donu text search capabilities to models list.

### Changes
- Add Donu `query-models` service
- Add model's text search query pill type
- Add Text Pill to models list search bar
- React to filter changes modifying model list

### Considerations
- Going back from Model Simulation view does not keep filters in search bar.
- Model querying is slow. This is on the Donu side and may be a result of inefficient regex search.
- Search results are not great when little metadata is present.

### Testing
- Regex search patterns can be used, ie. `*beta*`

### Artifacts:
Screen Shot 2021-08-05 at 11 36 11 AM

### DIFF
--- a/client/src/services/DonuService.ts
+++ b/client/src/services/DonuService.ts
@@ -75,6 +75,22 @@ const convertGrometEdgesDataToBGraphEdgesData = (edges: any[]): any[] => {
   return bgEdges;
 };
 
+/** Query available models from Donu API */
+export const queryDonuModels = async (text: string): Promise<any[]> => {
+  const request: Donu.Request = {
+    command: Donu.RequestCommand.QUERY_MODELS,
+    text,
+  };
+
+  const response = await callDonu(request);
+  if (response.status === Donu.ResponseStatus.success) {
+    const models = response.result ?? [];
+    return models;
+  } else {
+    console.error('[DONU Service] — queryDonuModels', response); // eslint-disable-line no-console
+  }
+};
+
 /** Fetch a complete list of available models from Donu API */
 export const fetchDonuModels = async (): Promise<Model.Model[]> => {
   const request: Donu.Request = {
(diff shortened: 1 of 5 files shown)


client/src/utils/QueryFieldsUtil.ts
```diff
@@ -219,6 +219,13 @@ const QUERY_FIELDS_MAP: QueryFieldMap = {
     baseType: 'string',
     lexType: 'string',
   },
+  // MODELS
+  MODELS_TEXT_SEARCH: {
+    ..._field('modelsTextSearch', 'Text Search'),
```
RosaRomeroGomez, line 224: I think this could be just `Keyword` to be consistent with how we do it in the Knowledge space. From Lex also says `Choose a field to search` so I think it makes more sense.

RosaRomeroGomez:
Just a comment. I think when we don't get results back, we should show `0 Computational models` in the counters to give some feedback to the user about a query with no results

client/src/services/DonuService.ts
```diff
@@ -75,6 +75,22 @@ const convertGrometEdgesDataToBGraphEdgesData = (edges: any[]): any[] => {
...
 };
 
+/** Query available models from Donu API */
+export const queryDonuModels = async (text: string): Promise<any[]> => {
+  const request: Donu.Request = {
+    command: Donu.RequestCommand.QUERY_MODELS,
+    text,
+  };
+
+  const response = await callDonu(request);
```
YohannParis, line 85: You might want to leverage `callDonuCache` on those expensive request

adamocarolli:
> Just a comment. I think when we don't get results back, we should show `0 Computational models` in the counters to give some feedback to the user about a query with no results

Done was a bug on the counter component.

PR #344, opened 2021-08-05
Add provenance graph skeleton

emilykuang:
**What?**
- Add the provenance graph prototype to the simulation view 
- Separating into multiple PRs: this is a UI skeleton to show interaction with the Open Provenance Graph button

**Testing**
- Open simulation view 
- Click on "Open Provenance Graph" and interact with the "Condensed | Expanded" buttons

**Screenrecording**
https://user-images.githubusercontent.com/23723378/128568745-a4b88b1b-79ba-4b42-9f44-5951657e0fe4.mov


client/src/components/widgets/ProvenanceCloseButton.vue
```diff
@@ -0,0 +1,43 @@
+<template>
```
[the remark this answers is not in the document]
emilykuang: The vertical positions are different between the two files, `CloseButton.vue` is for the drill-down panel so when I used this component, the button overlapped with the "Close Simulation" button. Maybe I can incorporate this as a button inside `ProvenanceGraph.vue` to avoid having an extra component?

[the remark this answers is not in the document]
RosaRomeroGomez: Actually, you should be able to use the `deep` selector to modify the css for that `CloseButton` component. https://www.telerik.com/blogs/understanding-vue-deep-css-selector The idea is that you might need to apply specific styling for a specific component depending on the parent or where you're using it. @YohannParis can probably explain this better or propose other options but I think this should work without having to create a new component.

client/src/views/Simulation/Simulation.vue
```diff
@@ -268,6 +293,30 @@
     onNodeHighlight (label: string): void {
       this.highlighted = label;
     }
+
+    onShowProvenanceGraph (): void {
```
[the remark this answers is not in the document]
emilykuang: Thanks for the suggestion, maybe a better name is `onToggleProvenanceGraph`? Since the Provenance Graph button acts as a toggle, we need to keep track of the current state of the graph. So when it is already open (true), the graph should close when the user clicks on the button. When it is closed (false), the graph should open when the user clicks on the button.

client/src/views/Simulation/Simulation.vue
```diff
@@ -37,6 +39,19 @@
       <search-bar />
     </aside>
 
+    <provenance-graph @close-pane="onCloseProvenanceGraph" :is-open="isOpenProvenanceGraph" @toggle-click="onProvenanceToggleClick">
+      <condensed-graph
+        v-if="provenanceActivePaneId === 'condensed'"
```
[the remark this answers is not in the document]
emilykuang: Sounds good, I have changed it so that there is only one component that watches `provenanceGraphData`.

client/src/components/widgets/ProvenanceCloseButton.vue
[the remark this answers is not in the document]
emilykuang: Thanks for the pointer, I have used this method to modify the position of the `CloseButton` and removed the new component

client/src/views/Simulation/components/ProvenanceGraph/ProvenanceGraph.vue
```diff
@@ -0,0 +1,87 @@
...
+    onClickExpanded (): void {
+      this.$emit('toggle-click', 'expanded');
+    }
+  }
+</script>
+
+<style scoped>
+.provenance-graph-container {
+  background-color: var(--bg-graphs);
+  box-shadow: 0 4px 8px 0 rgba(0,0,0,0.2);
```
YohannParis, line 59: I don't think we need the box shadow, the borders are good enough.
But I let that call be made by @RosaRomeroGomez

client/src/views/Simulation/components/ProvenanceGraph/ProvenanceGraph.vue
```diff
@@ -0,0 +1,87 @@
...
+          > Condensed
+        </button>
+        <button
+          class="btn btn-secondary"
+          @click="onClickExpanded"
+          > Expanded
+        </button>
+      </div>
+    </aside>
+    <close-button @close="onClose"/>
```
YohannParis, line 17: I don't thik this button is necessary as you can use the same button to Open and Close the provenance graph panel

client/src/views/Simulation/components/ProvenanceGraph/ProvenanceGraph.vue
```diff
@@ -0,0 +1,87 @@
+<template>
+  <div class="provenance-graph-container" v-if="isOpen">
+    <aside slot="left">
+      <div class="btn-group btn-group-sm">
```
YohannParis, line 4: Those would be better served as radio toggle:
https://getbootstrap.com/docs/4.3/components/buttons/#checkbox-and-radio-buttons

client/src/views/Simulation/components/ProvenanceGraph/ProvenanceGraph.vue
```diff
@@ -0,0 +1,87 @@
...
+  import CloseButton from '@/components/widgets/CloseButton.vue';
+
+  const components = {
+    CloseButton,
+  };
+
+  @Component({ components })
+  export default class ProvenanceGraph extends Vue {
+    @Prop({ default: false })
+    isOpen: boolean;
```
YohannParis, line 37: This doesn't need to be a prop if it's never used elsewhere.
The `v-if="isOpen"` can be handle by the parent component directly.

client/src/views/Simulation/Simulation.vue
```diff
@@ -10,9 +10,11 @@
         Search
       </button>
 
-      <button class="btn btn-primary">
+      <button
+        class="btn btn-primary"
+        @click="onToggleProvenanceGraph">
         <font-awesome-icon :icon="['fas', 'project-diagram' ]" />
-        Provenance Graph
+        <span>{{ isProvenanceGraphOpen ? 'Close Provenance Graph' : 'Open Provenance Graph' }}</span>
```
YohannParis, line 17: You can avoid repetition here, and make the code more succinct and readeable.
`<span>{{ isProvenanceGraphOpen ? 'Close' : 'Open' }}  Provenance Graph</span>`

YohannParis (approved):
Looks great !
All my comments are minors, mostly to help improve the code readability.

client/src/views/Simulation/components/ProvenanceGraph/ProvenanceGraph.vue
emilykuang: Thanks, I have removed it

client/src/views/Simulation/components/ProvenanceGraph/ProvenanceGraph.vue
emilykuang: We can keep it for now since it gives the user two ways to close the panel.

client/src/views/Simulation/components/ProvenanceGraph/ProvenanceGraph.vue
emilykuang: Thanks for the suggestion, I've made a note to come back to this as polishing changes after I load some data in.

client/src/views/Simulation/components/ProvenanceGraph/ProvenanceGraph.vue
emilykuang: Thanks, I've removed the prop.

RosaRomeroGomez (approved):
Thanks!

PR #348, opened 2021-08-06
Consolidated fixes

### LINKED ISSUE #343: Ability to search by DOI/ASKE-ID in the Knowledge space
We can create a new pill to search for in the docsCards view.

mj3cheun:
The following changes were made in this PR (sorry the branch name is misleading)

- In the right side knowledge panel of the models view, allow aske-ids from both code collection and text reference metadata types to be used in generating the knowledge list. If both metadata types exist, then the results will be combined
- In the right side knowledge panel of the models view, if a knowledge title is clicked, then instead of opening an external link to the knowledge source, open a modal identical to those used in the docCards view showing the knowledge result metadata
- In the left side metadata panel of the models view, modify the documents section so that instead of a dropdown providing more metadata, a modal containing the metadata is opened instead
- In the dosCards view, searching by aske-id is now permitted (Fixes #343)

To test: verify all this is correct

Notes:
- I notice that the stacking order is incorrect when the right side knowledge panel model is opened, the reason for this is explained in the following and can be resolved by removing the z-index from `.drilldown-panel-container`. Anyone got any reason why this is not safe to do?
- You will need to modify the code to test the right side knowledge panel because we arent getting metadata results from the current gromet. To do this, comment out line 376 in Model.vue.

### DIFF
--- a/client/src/components/Modals/ModalDocument.vue
+++ b/client/src/components/Modals/ModalDocument.vue
@@ -49,10 +49,10 @@
 
   import SimilarDocs from '@/components/SimilarDocs.vue';
   import Modal from '@/components/Modal.vue';
-  import { CardInterface } from '@/types/types';
   import {
     CosmosArtifactInterface,
     CosmosArtifactObjectInterface,
+    CosmosSearchObjectsInterface,
     CosmosSearchBibjsonInterface,
     CosmosSearchChildrenInterface,
   } from '@/types/typesCosmos';
@@ -65,7 +65,7 @@
 
   @Component({ components })
   export default class ModalDocument extends Vue {
-    @Prop({ required: false }) private data: CardInterface;
+    @Prop({ required: false }) private data: CosmosSearchObjectsInterface;
     @Prop({ required: false }) private artifact: CosmosArtifactInterface;
     @Prop({ default: false }) linkToKnowledgeSpace: boolean;
 
@@ -74,11 +74,11 @@
     }
 
     get bibjson (): CosmosSearchBibjsonInterface {
-      return this.artifact?.bibjson ?? this.data?.raw?.bibjson;
+      return this.artifact?.bibjson ?? this.data?.bibjson;
     }
 
     get children (): CosmosSearchChildrenInterface[] | CosmosArtifactObjectInterface[] {
-      return this.artifact?.objects ?? this.data?.raw.children ?? [];
+      return this.artifact?.objects ?? this.data?.children ?? [];
     }
 
     get doi (): string {
@@ -94,7 +94,7 @@
     }
 
     get title (): string {
-      return this.bibjson?.title ?? this.data?.title ?? '';
+      return this.bibjson?.title ?? '';
     }
 
     get excerpt (): string {
@@ -106,7 +106,7 @@
     }
 
     get imagePreview (): string {
-      return this.children?.[0]?.bytes ?? this.data?.previewImageSrc;
+      return this.children?.[0]?.bytes;
     }
 
     get imageStyle (): any {
--- a/client/src/components/Modals/ModalDocument.vue
+++ b/client/src/components/Modals/ModalDocument.vue
@@ -1,8 +1,15 @@
 <template>
   <modal @close="onClickClose">
-    <div slot="header">
-      <a :href="url" target="_blank"><h3>{{ title }}</h3></a>
-      <h5 class="m-0">{{ doi }}</h5>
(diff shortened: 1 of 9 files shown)


client/src/components/Modals/ModalDocument.vue
```diff
@@ -65,7 +65,7 @@
 
   @Component({ components })
   export default class ModalDocument extends Vue {
-    @Prop({ required: false }) private data: CardInterface;
+    @Prop({ required: false }) private data: CosmosSearchObjectsInterface;
```
YohannParis, line 68: I am loving it!

client/src/utils/QueryFieldsUtil.ts
```diff
@@ -194,6 +194,13 @@ const QUERY_FIELDS_MAP: QueryFieldMap = {
     lexType: 'string',
     order: 2,
   },
+  COSMOS_ASKE_ID: {
```
RosaRomeroGomez, line 197: Could we also allow search by DOI?

RosaRomeroGomez:
Some comments about this PR:

- For the right-side panel, I'm not sure which nodes or boxes have that actual metadata where I can test out aske-ids. I guess you're using the SIR GrFN gromet to test this out, right? You should include more concrete steps in the `testing` section. 
- For the left-side panel, after opening the modal dialog, on clicking the `Explore all in the knowledge space` link, I expected an automatic aske-id pill being created in the docsCards view so we close the workflow from Models to Knowledge.

image

RosaRomeroGomez:
> You will need to modify the code to test the right side knowledge panel because we arent getting metadata results from the current gromet. To do this, comment out line 376 in Model.vue.

Aren't we getting any metadata? For which models? I'm a bit confused.

client/src/views/Models/components/DrilldownPanel/KnowledgePane.vue
```diff
@@ -3,7 +3,7 @@
      <div v-if="!isEmptyData" class="mt-3 documents-container hide-scrollbar">
       <div class="mb-1 px-2 py-4 d-flex rounded-lg border" v-for="(object, index) in data.objects" :key="index">
         <div class="flex-grow-1">
-            <a :href="object.bibjson.link[0].url" target="_blank">
+            <a href="/" @click="e => openModal(e, index)">
```
YohannParis, line 6: Because this is not a link anymore, you should use a `<button type="button" class="btn btn-link" />` instead.

client/src/views/Models/components/MetadataPanel.vue
```diff
@@ -36,16 +36,11 @@
...
-            </summary>
-            <h6>File</h6>
-            <p><a :href="doc.bibjson.file_url">{{ doc.bibjson.file }}</a></p>
-            <h6>Author(s)</h6>
-            <p>{{ doc.bibjson.author.map(a => a.name) | ArrayToList }}</p>
-          </details>
+          <summary v-for="(doc, index) in datum.documents" :key="index" :title="doc.uid">
+            <a :href="doc.bibjson.website.url">{{ doc.bibjson.title }}</a>
+            {{ doc.global_reference_id.type }} {{ doc.global_reference_id.id }}
+            <div class="mt-3"><a href="/" @click="e => openModal(e, doc)">Show more...</a></div>
```
YohannParis, line 42: You should use a `<button type="button" class="btn btn-link" />` instead of a link, and you can apply the bootstrap classname `mt-3` to it to avoid the need of a single child `<div>`.

YohannParis (approved):
Apart of some small changes codewise, and Rosa comments, looks good to me.

client/src/components/Modals/ModalDocument.vue
```diff
@@ -1,8 +1,15 @@
...
-      <a :href="url" target="_blank"><h3>{{ title }}</h3></a>
-      <h5 class="m-0">{{ doi }}</h5>
+    <div class="flex-grow-1" slot="header">
+      <a :href="url" target="_blank">
+        <h3>{{ title }}</h3>
+      </a>
+      <div class="d-flex align-items-center justify-content-between">
+        <h5 class="m-0">{{ doi }}</h5>
+        <button class="btn btn-primary" @click="openKnowledgeView(true)">
+          Search Document Artifacts
```
RosaRomeroGomez, line 10: I think we probably want to say something like `Search Document in the context of the Knowledge space` and remove the link below.

RosaRomeroGomez (approved):

PR #352, opened 2021-08-10
Simulation legend

### LINKED ISSUE #291: Add legend to simulate page to clarify the colour encodings used by the panels

YohannParis:
**What**
- fix #291 

**How**
- Separated the new model legends, into a legend panel to contains legends and a model legend.
- Created a new simulation legend for the runs.

**Videos**

https://user-images.githubusercontent.com/636801/128936065-5ea672b1-a93f-403e-adcb-cb4ee433cd31.mov

### DIFF
--- a/client/src/views/Simulation/components/SimulationLegend.vue
+++ b/client/src/views/Simulation/components/SimulationLegend.vue
@@ -2,7 +2,7 @@
   <ul class="legend-simulation">
     <li class="run current">Current Run</li>
     <li class="run previous">Previous Runs</li>
-    <li class="run aggregate">Aggregate of all Runs</li>
+    <li class="run average">Average of all Runs</li>
     <li class="observational">Observational Data</li>
   </ul>
 </template>
@@ -34,18 +34,18 @@
     width: 3em;
   }
 
-  .aggregate::before {
+  .run.average::before {
     background-color: var(--selection);
   }
 
-  .current::before {
+  .run.current::before {
     background-color: var(--colours-nodes-default);
 
     /* To help separate from the panel background */
     border: 1px solid var(--nord4);
   }
 
-  .previous::before {
+  .run.previous::before {
     background-color: var(--colours-nodes-other);
   }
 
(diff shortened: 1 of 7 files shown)


client/src/views/Simulation/components/SimulationLegend.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+  <ul class="legend-simulation">
+    <li class="run current">Current Run</li>
+    <li class="run previous">Previous Runs</li>
+    <li class="run aggregate">Aggregate of all Runs</li>
```
RosaRomeroGomez, line 5: We probably want to add here the line for observational data (even if we don't have it yet). I think the color should be `#E75BCD` but @emilykuang can confirm.

RosaRomeroGomez (approved):
Good idea to just have everything in the same legend. Looks great.

client/src/views/Simulation/components/SimulationLegend.vue
YohannParis: Thanks, I was thimking that 3 legend buttons would be too much.

Do you think we need subtitles for the legend? I'm worriewd that user might think the runs colours are for edges. @emilykuang what are your thoughts?

emilykuang: Thanks for adding the observational data. Yes, the subtitles are good for differentiating between the Model Panel and Variable Panel, the divider looks great! 
I am just wondering about the wording for the yellow line: the meaning of `Aggregate` is "calculated by the combination of units; total" which may be confusing, many a better word is `Average`?

emilykuang (approved):

client/src/views/Simulation/components/SimulationLegend.vue
RosaRomeroGomez: 100%

YohannParis: Screen Shot 2021-08-11 at 09 30 05

PR #358, opened 2021-08-11
Move layout storage to model store to ensure layout state is preserved across views

mj3cheun:

PR #361, opened 2021-08-12
Add layout to comparison

YohannParis:
**What**
- add feature where the user can select _Modeling Framework_ and _Layout_ on the _Comparison_ view. 
https://docs.google.com/presentation/d/18XoZ94mPRouNm9EO-4LvDViouXy5U_xhBXZkq-FFRpI/edit?disco=AAAAN2GIv-s&usp=comment_email_discussion&usp_dm=false&ts=6113df0c

**How**
- based on #358 
- Move the list used to display _Modeling Framework_ and _Layout_ dropdown inside their respective type files.
- used the same compoment from the _Model_ view for the controlers.
- added an error message if a model do not have a specific _Modeling Framework_ to display.

**Video**

https://user-images.githubusercontent.com/636801/129243674-438a5574-d7bf-4bc4-b082-b8b843db0624.mov

**Improvement**
This controller clearly needs to be its own component in the future, or maybe once we merge the _Model_ and _Comparison_ views, this will not be an issue any longer.


emilykuang (approved):
Look good, especially the message for when a representation is not available for a model.

RosaRomeroGomez:
Really nice job. Thanks @YohannParis . I only have two comments that are more enhancements than anything:

- It seems the expand button is still enabled when we are in the comparison view. However, clicking on it doesn't really do anything. I think we should disable it when in this view.
image
- I feel that now we are a bit inconsistent with the placement of buttons when we select a single model vs multiple ones. I think we might want to move the modeling framework and layout dropdowns up for a single model as well. In this way we will be consistent. This design didn't really contemplate the possibility of multiple models but I think the hierarchy of information makes more sense that way.

RosaRomeroGomez (approved):

mj3cheun (approved):
looks good!

YohannParis:
> - It seems the expand button is still enabled when we are in the comparison view. However, clicking on it doesn't really do anything. I think we should disable it when in this view.

This has been addressed, I also removed the _Add/Remove All Parameters and Variables_.

> -  I feel that now we are a bit inconsistent with the placement of buttons when we select a single model vs multiple ones. I think we might want to move the modeling framework and layout dropdowns up for a single model as well. In this way we will be consistent. This design didn't really contemplate the possibility of multiple models but I think the hierarchy of information makes more sense that way.

The controls have been moved up top so both the _Model_ view and _Comparison_ view look identical.

RosaRomeroGomez (approved):
Many thanks!

PR #362, opened 2021-08-13
Refactored model comparison system and added support for multi-model comparison linked node highlighting

mj3cheun:
This PR accomplishes the following:

- Moves node selection from view files into the model store and created functions to calculate shared nodes and selected nodes given a specific model
- Adds highlight property to `EpiRenderer`
- Allows for (mostly) direct consumption of data provided by GTRI, Galois, etc. (Had to change the model names to match the model names fetched from donu)
- Supports one-to-many model comparison (demonstrated in the video)
- Full support for 2 model comparison with partial support for n model comparison (demonstrated in the video with rigged comparison input data) (full support requires model comparison data for each model to be compared and a function to combine multiple comparison data sources)

To test:
Click around the comparison view with the `SimpleSIR_metadata` and `SimpleChime+` functions

Notes:
- Right now the models view uses a different selection system (default node click highlight) compared to the comparison view (models store). This is not visible in the HMI but we might want to consider switching over the models view to use the new system later for consistency.
- Slightly concerned for efficiency, particularly with the `dataDecorationChanged` function

https://user-images.githubusercontent.com/14062458/129421371-785585d6-6c71-4abc-b5b9-bdd7edc3933f.mov

### DIFF
--- a/client/src/views/Comparison/Comparison.vue
+++ b/client/src/views/Comparison/Comparison.vue
@@ -82,7 +82,7 @@
         this.$route.params.model_id.split(',').forEach(this.setSelectedModels);
       }
 
-      return this.getModelsList.filter(model => this.getSelectedModelIds.map(Number).includes(model.id));
+      return this.getModelsList.filter(model => this.getSelectedModelIds.includes(model.id));
     }
 
     get gridMap (): string[][] {
--- a/client/src/views/Comparison/Comparison.vue
+++ b/client/src/views/Comparison/Comparison.vue
@@ -29,7 +29,6 @@
           class="comparison-panel"
           :key="index"
           :model="model"
-          :overlapping-elements="model.modelGraph[0].overlappingElements"
           :slot="('model_' + model.id)"
           :expandable="false"
         />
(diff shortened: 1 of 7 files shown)


RosaRomeroGomez:
Noticed something weird, it seems the stroke for nodes when in comparison mode, turns white instead of blue:

- Comparison mode:
image

- Single-model representation:
image

client/src/views/Comparison/Comparison.vue
```diff
@@ -92,18 +82,7 @@
...
-      selectedModels.forEach(model => {
-        if (model.name === 'SimpleSIR_metadata') {
-          nodes = Object.keys(MODEL_COMPARISON).map(node => ({ id: node }));
-        } else if (model.name === 'SimpleChime+') {
-          nodes = Object.values(MODEL_COMPARISON).map(node => ({ id: node }));
-        }
-        model.modelGraph[0].overlappingElements = { nodes, edges: [] };
-      });
-      return selectedModels;
+      return this.getModelsList.filter(model => this.getSelectedModelIds.map(Number).includes(model.id));
```
YohannParis, line 85: You don't need to convert to `Number`, `includes()` works on string with Numbers. 
i.e.
 ```JS
'1,2,3'.includes(3); // true
```

YohannParis (approved):
Looks good to me, and I love ❤️ the comments in the `utils/ModelUtil.ts`!!!

RosaRomeroGomez (approved):
Problem with the strokes got solved. Thanks!

PR #367, opened 2021-08-16
[FEATURE] Add document cluster queries

adamocarolli:
### Description
Add document title and DOI querying to document cluster view.

### Changes
- Add document title query field and query function
- Add document DOI query field and query function
- Add search pills to document cluster view.
- Add document title search pill to document cluster view.
- Add foreground/background doc cluster grafer settings to enable query highlighting.
- Send events between knowledge grafer controller and document cluster view to enable query highlighting. Events added are (1) foreground original data (2) background original data (3) update/add layers (4) remove layers (5) grafer loaded (6) loaded grafer layers

### Considerations
- The document cluster grafer layer/layer data is sent back to the document view to be stored in memory. This is because when a query is made that view will then filter down the data list to just what the query satisfies. The more efficient way of doing this is, instead of filtering existing layers, build up new query layer/layer data from scratch. However, this proved to be  more difficult than worthwhile for this pre-demo sprint.
- We are using fade in the grafer cluster view to differentiate cluster levels/heirarchy; here we are using desaturation to foreground/background the original grafer layers to differentiate original layers from query layers. Occasionally, especially when viewing sparsely populated sections of the ground it can be hard to differentiate between a desaturated node and a node that has been heavily faded. I've decided to keep it for this iteration to get a first pass in.
- The polygons around document nodes are not being highlighted on a query. Enabling this seems important, however, I've left it out of this initial pass as I had run low on time. 

### Testing
- No special instructions.

### Artifacts:
Screen Shot 2021-08-16 at 10 37 02 AM
https://user-images.githubusercontent.com/15199528/129582124-4768b006-00dc-4e6c-8bfb-8b28d7bbf5b7.mov

### DIFF
--- a/client/src/utils/QueryFieldsUtil.ts
+++ b/client/src/utils/QueryFieldsUtil.ts
@@ -95,9 +95,16 @@ const QUERY_FIELDS_MAP: QueryFieldMap = {
     baseType: 'integer',
     lexType: 'integer',
   },
-  DOCS_NODE_TITLE: {
-    ..._field('docsTitle', 'Document Title'),
-    ..._searchable('Document Title', false),
+  DOCS_CLUSTERS_NODE_TITLE: {
+    ..._field('docsClustersNodeTitle', 'Keyword'),
+    ..._searchable('Keyword', false),
+    baseType: 'string',
+    lexType: 'string',
+    order: 2,
+  },
+  DOCS_CLUSTERS_NODE_DOI: {
+    ..._field('docsClustersNodeDOI', 'DOI'),
+    ..._searchable('DOI', false),
     baseType: 'string',
     lexType: 'string',
   },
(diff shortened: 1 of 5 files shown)


client/src/utils/QueryFieldsUtil.ts
```diff
@@ -95,6 +95,12 @@ const QUERY_FIELDS_MAP: QueryFieldMap = {
     baseType: 'integer',
     lexType: 'integer',
   },
+  DOCS_NODE_TITLE: {
+    ..._field('docsTitle', 'Document Title'),
```
RosaRomeroGomez, line 99: We are really searching for a keyword included in the doc title, right? Wouldn't this be just called `Keyword` for keyword search?

RosaRomeroGomez (approved):

PR #369, opened 2021-08-16
Updated counters

### LINKED ISSUE #351: Change counters in Models view to be the same as in the Simulation Space
image
image

YohannParis:
**What**
- fix #351 

**How**
- Update counters on the `Model` and `Comparison` views to behave like in the `Simulation` view.
- Initialize simulation parameters and variables for each visible models

**Video**

https://user-images.githubusercontent.com/636801/129627883-950bc1ef-c4a2-4213-9f65-f175f0f3151d.mov

### DIFF
--- a/client/src/services/DonuService.ts
+++ b/client/src/services/DonuService.ts
@@ -219,8 +219,8 @@ export const fetchDonuModels = async (): Promise<Model.Model[]> => {
   }
 };
 
-/** Fetch the parameters of a model */
-export const getModelParameters = async (model: Model.Model, selectedModelGraphType: Model.GraphTypes): Promise<Donu.ModelParameter[]> => {
+/** Fetch the interface of a model */
+export const getModelInterface = async (model: Model.Model, selectedModelGraphType: Model.GraphTypes): Promise<Donu.ModelDefinition> => {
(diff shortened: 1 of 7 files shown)


RosaRomeroGomez:
My only concern with this (and I don't think it is that important) is that in the visualization we distinguish between initial conditions, variables, and parameters while donu only gives us parameters and variables. I think that's fine but we might need to clarify that if people ask.

RosaRomeroGomez (approved):

YohannParis:
> My only concern with this (and I don't think it is that important) is that in the visualization we distinguish between initial conditions, variables, and parameters while donu only gives us parameters and variables. I think that's fine but we might need to clarify that if people ask.

I can "easily" separate _initial conditions_ and _parameters_

emilykuang (approved):
Looks good!

YohannParis:
Here you go @RosaRomeroGomez 


https://user-images.githubusercontent.com/636801/129747477-c35f2951-1cf7-484e-a0ed-fa230e1d06f4.mov

RosaRomeroGomez (approved):
Lovely!Thank you so much.

PR #373, opened 2021-08-17
Update screenshots for models

RosaRomeroGomez:
**What**

- I don't think we'll get any new models so I updated the screenshots. 
NOTE: the rasmachine model doesn't seem to load at all, I included a placeholder screenshot there. 
image

**Testing**

- Open the Models view.


YohannParis (approved):
I'm not a fan of importing static assets in JS, but right now it's not the time to talk about it.

PR #378, opened 2021-08-17
Fix incorrect graph type crash in models view

mj3cheun:
Fixes crash when attempting to open `CHIME_SIR` from a fresh state due to incorrect assumed graph type.

To test, from a fresh state, opening `SimpleSIR_metadata`, `CHIME_SIR`, and `SimpleChime+` model views should not result in any errors (either visible or in console). The correct graph type should be used for `CHIME_SIR` and `SimpleChime+`. Graph type switching should be unaffected in the `SimpleSIR_metadata` model.


RosaRomeroGomez (approved):
Thanks!

YohannParis (approved):

PR #379, opened 2021-08-17
fix error message with marm model

YohannParis:
1. Updated how the parameter panel detect if a value is invalid or not.
2. Updated how the `Run` button is disabled based on the above.


RosaRomeroGomez:
Something I noticed now is that when we try to run it with some default parameters, we get the following error in donu after clicking on `run`. It seems like we are not sending the parameter values to donu (reading the error we get):

image

YohannParis:
> Something I noticed now is that when we try to run it with some default parameters, we get the following error in donu after clicking on `run`. It seems like we are not sending the parameter values to donu (reading the error we get):

This error was due to the new counter on the model panel that initialize the simulation model information twice... Now it works as it should, good catch on that one!


https://user-images.githubusercontent.com/636801/129993335-eeaba3ae-c6c9-4081-87fb-d317761d23d9.mov

RosaRomeroGomez (approved):
Lovely!

PR #385, opened 2021-08-18
Add more data to provenance graph

emilykuang:
**What?**
- Add more mock data and tooltips to provenance graph 
- Add session metadata header 

**How?**
- Use `expandCollapse` to expand and collapse containers and use `showTooltip` in SVGUtils.ts (_Note: there is a known issue with the tooltips where it takes the parent’s location when you have nested nodes_)

**Testing:**
- Open Simulation view with single and multiple models then play with the provenance graph 

Screen-recording: 

https://user-images.githubusercontent.com/23723378/129967909-c8958869-6470-40d4-90db-e9cf6abf0e7a.mov


client/src/views/Simulation/components/ProvenanceGraph/ProvenanceGraph.vue
```diff
@@ -12,6 +12,7 @@
           {{ layout.name }}
         </button>
       </div>
+      <h7>{{sessionMetadata}}</h7>
```
YohannParis, line 15: `<h7>` does not exist, `<h6>` is the limit.

YohannParis (approved):

client/src/views/Simulation/components/ProvenanceGraph/ProvenanceGraph.vue
RosaRomeroGomez: Yep, I'm getting an error.

RosaRomeroGomez (commented):
What's the main difference now between the collapsed/expanded modes when we have nested operations? I'd expect that by clicking on the `expanded` version, all the collapsed nodes get expanded. However, that's currently not the case.

emilykuang:
> What's the main difference now between the collapsed/expanded modes when we have nested operations? I'd expect that by clicking on the `expanded` version, all the collapsed nodes get expanded. However, that's currently not the case.

The difference between the **condensed** and **expanded** modes are the types of nodes displayed. The condensed mode only has the operation nodes (rectangles for select, edit, execute) whereas the expanded mode also includes the data nodes (ellipses for models, parameters, variables) in addition to the operations. The collapse/expand is for the container boxes which is just a visual change. Maybe we can rename "Expanded" to "Detailed" or something to be more clear?

client/src/views/Simulation/components/ProvenanceGraph/ProvenanceGraph.vue
emilykuang: Thanks, I've fixed it :)

RosaRomeroGomez:
> > What's the main difference now between the collapsed/expanded modes when we have nested operations? I'd expect that by clicking on the `expanded` version, all the collapsed nodes get expanded. However, that's currently not the case.
> 
> The difference between the **condensed** and **expanded** modes are the types of nodes displayed. The condensed mode only has the operation nodes (rectangles for select, edit, execute) whereas the expanded mode also includes the data nodes (ellipses for models, parameters, variables) in addition to the operations. The collapse/expand is for the container boxes which is just a visual change. Maybe we can rename "Expanded" to "Detailed" or something to be more clear?

Yes, I think that's a good idea. We have talked about this but since I forgot, that name confused me a bit.

emilykuang:
> > > What's the main difference now between the collapsed/expanded modes when we have nested operations? I'd expect that by clicking on the `expanded` version, all the collapsed nodes get expanded. However, that's currently not the case.
> > 
> > 
> > The difference between the **condensed** and **expanded** modes are the types of nodes displayed. The condensed mode only has the operation nodes (rectangles for select, edit, execute) whereas the expanded mode also includes the data nodes (ellipses for models, parameters, variables) in addition to the operations. The collapse/expand is for the container boxes which is just a visual change. Maybe we can rename "Expanded" to "Detailed" or something to be more clear?
> 
> Yes, I think that's a good idea. We have talked about this but since I forgot, that name confused me a bit.

Sounds good, I've renamed "expanded" to "detailed"

RosaRomeroGomez:
Thanks!

PR #386, opened 2021-08-18
Execute Functional Network

YohannParis:
**What**
- Enable the execution of _Functional Network_ models.

**How**
- Find which _Parameters_ and _Variables_ can be used for the simulation. 
    For example: as of now for _SimpleSIR_metadata_ only `'P:sir.out.S', 'P:sir.out.I', P:sir.out.R'` can be used as _Variables_ and `'P:sir.in.dt'` is used as _time_.
- Create special cases on the `DonuServices` to handle _SimpleSIR_metadata_ and _CHIME by hand_ and massage their metadata name to make them work with the model graph.
- Update `DonuTypes` to includes the difference between simulation of _Functional Network_ and a _Petri Net Classic_.
- Displayed the _Domain Parameter_ parameter differently to let the user understand why it's not editable.

**Other small fixes**
- Use UID for _Variables_ without a name.
- Update _Modeling Framework_ controls to close after selecting an option.
- Update the Simulation config to run 100 steps of 1.
- Update issue when resetting the Simulation the Parameters values would not reset.

**Video**

https://user-images.githubusercontent.com/636801/130145199-d6c94fd3-198d-4fea-bb10-4ea0c46c5358.mov

**Testing**
So far you can only run **SimpleSIR_metadata**.


client/src/services/DonuService.ts
```diff
@@ -193,9 +192,31 @@ export const getModelInterface = async (model: Model.Model, selectedModelGraphTy
...
+       * The DONU API needs to be clearer on which outputs can be selected programatically.
+       * Also, the HMI needs to be updated to allow the user to select which paramaters
+       * to use as `domain_parameter` for the simulation steps.
+       */
+      if (selectedModelGraphType === Model.GraphTypes.FunctionNetwork) {
+        const result = response?.result as Donu.ModelDefinition;
+        let outputs = [];
+        let domainParameter;
+        if (modelGraph.model === 'SimpleSIR_metadata_gromet_FunctionNetwork.json') {
+          outputs = ['P:sir.out.S', 'P:sir.out.I', 'P:sir.out.R'];
```
RosaRomeroGomez, line 206: Could we just have `S, I, R` for these variables?

client/src/types/typesDonu.ts
```diff
@@ -89,13 +91,15 @@ type Request = {
...
-  // 'dest-type'?: string;
+  outputs?: string[],
+  domain_parameter?: string,
   end?: number;
-  // name?: string;
   start?: number;
   step?: number;
   'sim-type'?: SimulationType | void,
   text?: string,
+  // name?: string;
```
RosaRomeroGomez, line 101: Can't we just remove this?

RosaRomeroGomez:
Something I noticed is that if I try to add a node from the Model Panel (from the visualization) to the Parameters Panel doesn't work anymore? It works for PNCs though.

client/src/services/DonuService.ts
YohannParis: I can hard code in the backend to rename them visually if that's what you are saying.
Those are `uid` and I cannot change them, but I can set their `metadata.name` based on what you want.

YohannParis:
> Something I noticed is that if I try to add a node from the Model Panel (from the visualization) to the Parameters Panel doesn't work anymore? It works for PNCs though.

I do not understand what you are talking about? You can drag-and-drop node to the drilldown panel?

client/src/services/DonuService.ts
RosaRomeroGomez: Yep, that is what I meant.

RosaRomeroGomez:
image
I think changing the names is now provoking this error ^?

YohannParis:
> I think changing the names is now provoking this error ^?
You find the weirdest bugs, how did you got this?

YohannParis:
> I think changing the names is now provoking this error ^?

Thanks for the catch up, this was a big bug that creeped into the store.
It has been solved. Checked the video on the top to see resolve.

RosaRomeroGomez:
Looks good to go to me!

RosaRomeroGomez (approved):

PR #394, opened 2021-08-19
Added Covid Emma Gromet

RosaRomeroGomez:
**What**

- Added aggregated version of the COVID-19 emmaa gromet to the models list
- Added screenshot for the models list view's cards.

image


YohannParis (approved):
Looks great!
It's almost if we could have kept our mocked data from last month ;)

PR #400, opened 2021-08-20
Parameter plot on model drilldown panel

### LINKED ISSUE #382: Include parameter plot under the Parameters tab in the Model drill-down panel
- Back in February, we used some xdd extracted data to populate the Parameters tab in the Model drill-down panel to show some values extracted from documents.
- The extracted data comes from this table: http://teststrata.geology.wisc.edu/xdd/extract.php?doc_set=xdd-covid-19 (we only used two columns (variable and local, then I used date for the tooltip)
- This is how it looked like back in February:
image

This is a rough sketch of the idea I think we could quickly implement this time. We could enable a dropdown menu under the Parameters tab that allows the user to explore existing data  in that table. We can show some variable by default (e.g. R0) but allow the user to select others, which updates the scatterplot at the bottom.
image

YohannParis:
**What**
- fix #382 

**How**
- Creating a basic JSON of the website data.
- Updating the scatter plot and parameters pane
- adding a dropdown to select which variables to display
- fix the tooltip and click on a dot to open COSMOS document
- Display the _Related Parameters_ back

**Video**

https://user-images.githubusercontent.com/636801/130286612-3ab2e41f-757f-4f6c-b362-3d7780d4e3e4.mov

**Consideration**
- The location axis is wonky and does not align well with the data.
- the zero axes sometimes is visible out of range
- the graph would improve drastically with a serious cleanup of the data.


RosaRomeroGomez:
I noticed the following for the drill-down panel. It seems that the width of the drill-down panel when the Params panel is open is larger than for the other tabs. I think we should actually make the other tabs look as wide as this one.

https://user-images.githubusercontent.com/10552785/130360617-8ab7a6a5-ba42-4766-89f5-26dc53ea6596.mov

RosaRomeroGomez (commented):
I left a comment there about the width of this panel but other than that, it looks good to me.

YohannParis:
I added more CSS properties to make sure the graph stay contained within the dimensions of the drilldown panel.
I also made the drilldown panel in the _Model_ view wider.

@RosaRomeroGomez feel free to change that value if it's too big.
And if you are happy about the changes, please go ahead and merge the PR ;)

RosaRomeroGomez (approved):

mj3cheun (approved):

PR #414, opened 2021-08-25
Graph centering

RosaRomeroGomez:
**What**
- For the demo, we weren't really showing nested graphs so I disabled the automatic collapsing of boxes. However, CHIME models are highly nested so we required to put that back. Collapsing by default was the cause of the graph disappearing. 
- Addressed this issue by implementing a function that centers the graph.
- Put back the automatic collapsing
- Extra:  I also added the ability to drag nodes around that was already available in the SVG lib (although this is not perfect and it doesn't work so well with collapse and expand).
**Testing**

- Open the current CHIME SIR base model
- It should collapse top level boxes by default and get centered properly.

https://user-images.githubusercontent.com/10552785/130765533-95a3aba7-62cc-4be9-bfcb-d4b99ea7b856.mov

### DIFF
--- a/client/src/graphs/svg/renderers/EpiRenderer.ts
+++ b/client/src/graphs/svg/renderers/EpiRenderer.ts
@@ -128,6 +128,15 @@ export default class EpiRenderer extends SVGRenderer {
       .attr('color', 'SourceGraphic');
   }
 
+  centerGraph () {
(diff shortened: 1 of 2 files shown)


adamocarolli (approved):
Looks good!

